PR #7, opened 2022-12-14
feat: Beefy vaults

karelianpie:
⚠️ WIP Add beefy vaults


karelianpie:
@Majorfi was this merged into [yvempire](https://github.com/yearn/yearn.fi/tree/yvempire)? If so, we could close this

PR #8, opened 2022-12-16
feat: Add memory to the search bar

karelianpie:
## Description

* Add context so that search and category filters are preserved on page change.

## Screen recording

https://user-images.githubusercontent.com/78794805/208098701-569ad1a1-5cba-455c-93c0-15e86d5c9377.mov


karelianpie:
It could use the upcoming `useSessionStorage` hook from https://github.com/yearn/web-lib/pull/142

PR #12, opened 2022-12-19
feat: Pattern for multiple external service support

karelianpie:
## Description

Add structure for supporting multiple external services, namely [wido](https://www.joinwido.com/), [portals.fi](https://portals.fi/) and [CoW](https://swap.cow.fi/#/1/swap/WETH).

## Related Issue

N/A

## Motivation and Context

Use multiple external services for related actions

## How Has This Been Tested?

Ran locally

### DIFF
--- a/apps/common/hooks/useExternalQuote.ts
+++ b/apps/common/hooks/useExternalQuote.ts
@@ -1,4 +1,4 @@
-import {quote as widoQuote} from 'wido';
+import {quote} from 'quote';
 
 import type {QuoteRequest as TWidoRequest, QuoteResult as TWidoResult} from 'wido';
 
@@ -21,9 +21,13 @@ type TPortalsResult = {
 
 export type TPortalsQuote = { req: { type: 'portals'; request: TPortalsRequest }; res: TPortalsResult };
 
+// TODO
+export type TCowQuote = { req: { type: 'cow'; request: unknown }; res: unknown };
+
 type TExternalService<T> =
     T extends 'wido' ? TWidoQuote :
     T extends 'portals' ? TPortalsQuote :
+    T extends 'cow' ? TCowQuote :
 	never;
 
 export async function useExternalQuote<T>({type, request}: TExternalService<T>['req']): Promise<TExternalService<T>['res']> {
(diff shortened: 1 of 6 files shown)


apps/common/hooks/useExternalQuote.ts
```diff
@@ -0,0 +1,148 @@
...
+// 	} else if (type === 'portals') {
+// 		return response2;
+// 	}
+// 	throw new Error('Function not implemented.');
+// }
+
+export function useExternalServiceQuote<T>({type, request}: TExternalService<T>['req']): TExternalService<T>['res'] {
+	const [widoQuote, getWidoQuote] = useCowQuote()/*useWidoQuote()*/;
+	const [portalsQuote, getPortalsQuote] = useCowQuote()/*usePortalsQuote()*/;
+	const [cowQuote, getCowQuote] = useCowQuote();
```
karelianpie, line 122: As we're not really making use of these individually, we could just assign to a single variable, it will make the code a bit easier to read
```suggestion
	const widoQuote = useCowQuote()/*useWidoQuote()*/;
	const portalsQuote = useCowQuote()/*usePortalsQuote()*/;
	const cowQuote = useCowQuote();
```

apps/common/hooks/useExternalQuote.ts
```diff
@@ -0,0 +1,148 @@
...
+
+	const quoteMapping = useMemo((): TDict<[TExternalService<T>['res'], CallableFunction]> => ({
+		'wido': [widoQuote, getWidoQuote],
+		'portals': [portalsQuote, getPortalsQuote],
+		'cowswap': [cowQuote, getCowQuote]
+	}), [cowQuote, getCowQuote, getPortalsQuote, getWidoQuote, portalsQuote, widoQuote]);
```
karelianpie, line 128:
```suggestion
	const quoteMapping = useMemo((): TDict<[TExternalService<T>['res'], CallableFunction]> => ({
		'wido': widoQuote,
		'portals': portalsQuote,
		'cowswap': cowQuote
	}), [cowQuote, portalsQuote, widoQuote]);
```

apps/common/hooks/useExternalQuote.ts
```diff
@@ -0,0 +1,148 @@
...
+
+		const	[, fetcher] = quoteMapping[type];
+		if (fetcher) {
+			fetcher(parsedRequest);
+		} else {
+			throw new Error(`Unknown service ${type}`);
+		}
+	}, [stringifiedRequest, type]);
+
+	return quoteMapping[type][0];
```
karelianpie, line 147: I would make this a normal object rather than an array, then we could `quoteMapping[type].quote`

apps/common/hooks/useExternalQuote.ts
```diff
@@ -0,0 +1,148 @@
...
+	const stringifiedRequest = JSON.stringify(request);
+	useEffect((): void => {
+		const	parsedRequest = JSON.parse(stringifiedRequest, (_key: string, value: T & { type?: string}): T | BigNumber => {
+			if (value?.type === 'BigNumber') {
+				return BigNumber.from(value);
+			}
+			return value;
+		});
+
+		const	[, fetcher] = quoteMapping[type];
```
karelianpie, line 139: https://github.com/yearn/yearn.fi/pull/12/files#r1052433478 would make this more readable too

const {fetcher} = quoteMapping[type];

apps/common/hooks/useExternalQuote.ts
```diff
@@ -0,0 +1,148 @@
...
+// 	}
+// 	throw new Error('Function not implemented.');
+// }
+
+export function useExternalServiceQuote<T>({type, request}: TExternalService<T>['req']): TExternalService<T>['res'] {
+	const [widoQuote, getWidoQuote] = useCowQuote()/*useWidoQuote()*/;
+	const [portalsQuote, getPortalsQuote] = useCowQuote()/*usePortalsQuote()*/;
+	const [cowQuote, getCowQuote] = useCowQuote();
+
+	const quoteMapping = useMemo((): TDict<[TExternalService<T>['res'], CallableFunction]> => ({
```
karelianpie, line 124: Also, I think this mapping only works because we are only using the cow hook. Once we add more than types won't be compatible

PR #13, opened 2022-12-20
Feat/external service cow

Majorfi:
## Basic stuffs:
- `apps/common/components/ListHero.tsx` simple hydratation warning suppress to avoir issue in dev mode. No impact otherwise
- `apps/common/hooks/useAllowanceFetcher.tsx` creation of a react hook to get the allowance for a given `inputERC20` for a given `spender`. This only returns a callback function that can be called when this allowance is required
- `apps/vaults/hooks/useVaultEstimateOutFetcher.tsx` creation of a react hook to get the estimate out for a given vault. This only returns a callback function that can be called when this allowance is required
- `apps/common/types/types.tsx` add `solveVia` to the `TDropdownOption` type to indicate which solver to use with this options. Later, this may replace `zapVia`, but not yet.
- `apps/common/utils/index.tsx` add a basic `DefaultTNormalizedBN` to save troubles
- `apps/vaults/Wrapper.tsx` add a wrapper around `WalletForZapApp`

## Code clean
- Split QuickAction component in multiple files:
  - `apps/vaults/components/details/actions/QuickActionsTo.tsx`
  - `apps/vaults/components/details/actions/QuickActionsSwitch.tsx`
  - `apps/vaults/components/details/actions/QuickActionsFrom.tsx`
  - `apps/vaults/components/details/actions/QuickActionsButtons.tsx`
- Move VaultTabsDetails in a separate folder:
  - `apps/vaults/components/details/tabs/VaultDetailsAbout.tsx`
  - `apps/vaults/components/details/tabs/VaultDetailsHistorical.tsx`
  - `apps/vaults/components/details/tabs/VaultDetailsStrategies.tsx`

## Solvers
The idea was to find a way to make the app scalable with multiple solvers. A solver, or service, or flow, is a group of actions that could lead to a determined result. Our solvers are all exporting the same type of actions:
- `expectedOut` -> For a given deposit/withdraw/zap action, a given tokenIn and a given amount, returns, as `TNormalizedBN` the estimate out the user will receive
- `isLoadingExpectedOut` -> Indicate if the request for this estimate out is pending result
- `onApprove` -> Action to realize in order to approve our inputToken to the solver 
- `onExecuteDeposit` -> Action to realize in order to deposit our inputToken and receive our outputToken
- `onExecuteWithdraw` -> Action to realize in order to withdraw our inputToken and receive our outputToken

The Solvers right now are:
- `Vanilla` solver, which is used to deposit/withdraw directly to/from the vault.
- `PartnerContract` solver, which is used to deposit to a vault from our partnerContract, and withdraw directly from the vault.
- `ChainCoin` solver, which is used to deposit the chain native Coin (ETH/FTM/etc.) to the corresponding Yearn Wrapped Vault, and to withdraw Yearn Wrapped Vault Token to the chain native Coing directly.
- `Cowswap` solver, which is used to perform a "zap" in action to deposit any supported tokens to the vault.
- `Wido` solver, which is used to perform a "zap" in action to deposit any supported tokens to the vault. _TODO_
- `Portals` solver, which is used to perform a "zap" in action to deposit any supported tokens to the vault. _TODO_

## Flow
When loading the vault page, the `useActionFlow` context will be loaded. This context handle all the stuff related to our app and the deposit/withdraw flow: handle the inputs, the deposit/withdraw status etc. It is also responsible of determining which solver to use, based on the input and the output token. Ex:
- If input or output is Eth, use `chainCoin` 
- If input has the `SolveVia` property set to `cowswap`, use `cowswap`
- If partnerContract usage is disabled use `vanilla`
- Otherwise use `partnerContract`
This is where we will load the zaps options. _TODO_

Then, under this context, another context is loaded, the `useSolver` context. This one will load the solvers specific hooks and, based on the solver to user, returns a standardized schema to use:
```ts
export type TWithSolver = {
  currentSolver: Solvers;
  expectedOut: TNormalizedBN;
  isLoadingExpectedOut: boolean;
  onApprove: () => Promise<void>;
  onExecuteDeposit: () => Promise<void>;
  onExecuteWithdraw: () => Promise<void>;
}
```

Then we can use theses elements to solve our deposit/withdraw/zap

### DIFF
--- a/apps/vaults/contexts/useActionFlow.tsx
+++ b/apps/vaults/contexts/useActionFlow.tsx
@@ -1,5 +1,4 @@
 import React, {createContext, useCallback, useContext, useEffect, useMemo, useReducer, useState} from 'react';
-import {ethers} from 'ethers';
 import {isSolverDisabled, Solver} from '@vaults/contexts/useSolver';
 import {useWalletForZap} from '@vaults/contexts/useWalletForZaps';
 import {setZapOption} from '@vaults/utils/zapOptions';
@@ -111,8 +110,8 @@ function ActionFlowContextApp({children, currentVault}: {children: ReactNode, cu
 	const [isDepositing, isUsingPartnerContract] = useContextualIs(actionParams?.selectedOptionTo, currentVault);
 
 	const maxDepositPossible = useMemo((): TNormalizedBN => {
-		const	vaultDepositLimit = formatBN(currentVault.details.depositLimit) || ethers.constants.Zero;
-		const	userBalance = balances?.[toAddress(actionParams?.selectedOptionFrom?.value)]?.raw || ethers.constants.Zero;
+		const	vaultDepositLimit = formatBN(currentVault?.details?.depositLimit);
+		const	userBalance = formatBN(balances?.[toAddress(actionParams?.selectedOptionFrom?.value)]?.raw);
 		if (actionParams?.selectedOptionFrom?.value === currentVault?.token?.address && isDepositing) {
 			if (userBalance.gt(vaultDepositLimit)) {
 				return (toNormalizedBN(vaultDepositLimit, currentVault.token.decimals));
@@ -284,10 +283,10 @@ function ActionFlowContextApp({children, currentVault}: {children: ReactNode, cu
 	** If not, the amount is set to the user balance for that token.
 	**********************************************************************************************/
 	const	updateParams = useCallback((_selectedFrom: TDropdownOption, _selectedTo: TDropdownOption): void => {
-		const	userBalance = balances?.[toAddress(_selectedFrom?.value)]?.raw || ethers.constants.Zero;
+		const	userBalance = formatBN(balances?.[toAddress(_selectedFrom?.value)]?.raw);
 		let	_amount = toNormalizedBN(userBalance, _selectedFrom?.decimals || currentVault?.token?.decimals || 18);
 		if (isDepositing) {
-			const	vaultDepositLimit = formatBN(currentVault.details.depositLimit) || ethers.constants.Zero;
+			const	vaultDepositLimit = formatBN(currentVault?.details?.depositLimit);
 			if (_selectedFrom?.value === currentVault?.token?.address) {
 				if (userBalance.gt(vaultDepositLimit)) {
 					_amount = toNormalizedBN(vaultDepositLimit, currentVault.token.decimals);
--- a/apps/vaults/hooks/useSolverCowswap.tsx
+++ b/apps/vaults/hooks/useSolverCowswap.tsx
@@ -7,11 +7,11 @@ import {isSolverDisabled, Solver} from '@vaults/contexts/useSolver';
 import {yToast} from '@yearn-finance/web-lib/components/yToast';
 import {useWeb3} from '@yearn-finance/web-lib/contexts/useWeb3';
 import {isZeroAddress, toAddress} from '@yearn-finance/web-lib/utils/address';
+import {SOLVER_COW_VAULT_RELAYER_ADDRESS} from '@yearn-finance/web-lib/utils/constants';
 import {formatBN, toNormalizedBN} from '@yearn-finance/web-lib/utils/format.bigNumber';
 import {Transaction} from '@yearn-finance/web-lib/utils/web3/transaction';
 import {useYearn} from '@common/contexts/useYearn';
 import {approvedERC20Amount, approveERC20, isApprovedERC20} from '@common/utils/actions/approveToken';
-import {COW_VAULT_RELAYER_ADDRESS} from '@common/utils/constants';
 
 import type {AxiosError} from 'axios';
 import type {TTxStatus} from '@yearn-finance/web-lib/utils/web3/transaction';
@@ -277,7 +277,7 @@ export function useSolverCowswap(): TSolverContext {
 		const allowance = await approvedERC20Amount(
 			provider as ethers.providers.Web3Provider,
 			toAddress(request.current.inputToken.value), //Input token
-			toAddress(COW_VAULT_RELAYER_ADDRESS) //Spender, aka Cowswap solver
+			toAddress(SOLVER_COW_VAULT_RELAYER_ADDRESS) //Spender, aka Cowswap solver
 		);
 		return toNormalizedBN(allowance, request.current.inputToken.decimals);
 	}, [provider, request]);
@@ -299,7 +299,7 @@ export function useSolverCowswap(): TSolverContext {
 		const	isApproved = await isApprovedERC20(
 			provider as ethers.providers.Web3Provider,
 			toAddress(request.current.inputToken.value), //token to approve
-			toAddress(COW_VAULT_RELAYER_ADDRESS), //Cowswap relayer
+			toAddress(SOLVER_COW_VAULT_RELAYER_ADDRESS), //Cowswap relayer
 			amount
 		);
 		if (isApproved) {
@@ -310,7 +310,7 @@ export function useSolverCowswap(): TSolverContext {
 		new Transaction(provider, approveERC20, txStatusSetter)
 			.populate(
 				toAddress(request.current.inputToken.value), //token to approve
-				toAddress(COW_VAULT_RELAYER_ADDRESS), //Cowswap relayer
+				toAddress(SOLVER_COW_VAULT_RELAYER_ADDRESS), //Cowswap relayer
 				amount
 			)
 			.onSuccess(async (): Promise<void> => {
--- a/apps/common/components/BalanceReminderPopover.tsx
+++ b/apps/common/components/BalanceReminderPopover.tsx
@@ -1,6 +1,5 @@
 import React, {Fragment, useMemo} from 'react';
 import Image from 'next/image';
-import {ethers} from 'ethers';
 import {Popover, Transition} from '@headlessui/react';
 import {useWeb3} from '@yearn-finance/web-lib/contexts/useWeb3';
 import {useChainID} from '@yearn-finance/web-lib/hooks/useChainID';
@@ -8,6 +7,7 @@ import IconAddToMetamask from '@yearn-finance/web-lib/icons/IconAddToMetamask';
 import IconCross from '@yearn-finance/web-lib/icons/IconCross';
 import IconWallet from '@yearn-finance/web-lib/icons/IconWallet';
 import {toAddress, truncateHex} from '@yearn-finance/web-lib/utils/address';
+import {formatBN} from '@yearn-finance/web-lib/utils/format.bigNumber';
 import {formatAmount} from '@yearn-finance/web-lib/utils/format.number';
 import {useWallet} from '@common/contexts/useWallet';
 import {useYearn} from '@common/contexts/useYearn';
@@ -93,7 +93,7 @@ export default function BalanceReminderPopover(): ReactElement {
 
 	const	nonNullBalances = useMemo((): TDict<TBalanceData> => {
 		const	nonNullBalances = Object.entries(balances).reduce((acc, [address, balance]): TDict<TBalanceData> => {
-			if (!(balance?.raw || ethers.constants.Zero).isZero()) {
+			if (!formatBN(balance?.raw).isZero()) {
 				acc[toAddress(address)] = balance;
 			}
 			return acc;
--- a/apps/common/contexts/useYearn.tsx
+++ b/apps/common/contexts/useYearn.tsx
@@ -1,12 +1,11 @@
 import React, {createContext, memo, useContext, useMemo} from 'react';
-import {ethers} from 'ethers';
 import useSWR from 'swr';
 import {Solver} from '@vaults/contexts/useSolver';
 import {useSettings} from '@yearn-finance/web-lib/contexts/useSettings';
 import {useWeb3} from '@yearn-finance/web-lib/contexts/useWeb3';
 import {useLocalStorage} from '@yearn-finance/web-lib/hooks';
 import {useChainID} from '@yearn-finance/web-lib/hooks/useChainID';
-import {toAddress} from '@yearn-finance/web-lib/utils/address';
+import {addressZero, toAddress} from '@yearn-finance/web-lib/utils/address';
 import {baseFetcher} from '@yearn-finance/web-lib/utils/fetchers';
 
 import type {ReactElement} from 'react';
@@ -37,7 +36,7 @@ const	defaultProps: TYearnContext = {
 	},
 	prices: {},
 	tokens: {},
-	vaults: {[ethers.constants.AddressZero]: undefined},
+	vaults: {[addressZero]: undefined},
 	isLoadingVaultList: false,
 	mutateVaultList: async (): Promise<void> => Promise.resolve(),
 	zapSlippage: 0.1,
--- a/apps/common/hooks/useAllowanceFetcher.tsx
+++ b/apps/common/hooks/useAllowanceFetcher.tsx
@@ -2,7 +2,7 @@ import {useCallback} from 'react';
 import {ethers} from 'ethers';
 import {useWeb3} from '@yearn-finance/web-lib/contexts/useWeb3';
 import {toAddress} from '@yearn-finance/web-lib/utils/address';
-import {toNormalizedBN} from '@yearn-finance/web-lib/utils/format.bigNumber';
+import {formatBN, toNormalizedBN} from '@yearn-finance/web-lib/utils/format.bigNumber';
 
 import type {TAddress} from '@yearn-finance/web-lib/utils/address';
 import type {TDropdownOption, TNormalizedBN} from '@common/types/types';
@@ -29,7 +29,7 @@ export function	useAllowanceFetcher(): (args: TAllowanceFetcher) => Promise<TNor
 		const	address = await (provider as ethers.providers.Web3Provider).getSigner().getAddress();
 
 		try {
-			const	tokenAllowance = await contract.allowance(address, spender) || ethers.constants.Zero;
+			const	tokenAllowance = formatBN(await contract.allowance(address, spender));
 			const	effectiveAllowance = toNormalizedBN(tokenAllowance, inputToken.decimals);
 			return effectiveAllowance;
 		} catch (error) {
--- a/apps/common/hooks/useBalances.tsx
+++ b/apps/common/hooks/useBalances.tsx
@@ -1,16 +1,16 @@
 import {useCallback, useEffect, useMemo, useRef, useState} from 'react';
 import {Contract} from 'ethcall';
-import {ethers} from 'ethers';
 import {useWeb3} from '@yearn-finance/web-lib/contexts/useWeb3';
 import {useChainID} from '@yearn-finance/web-lib/hooks/useChainID';
 import ERC20_ABI from '@yearn-finance/web-lib/utils/abi/erc20.abi';
 import {toAddress} from '@yearn-finance/web-lib/utils/address';
 import {ETH_TOKEN_ADDRESS, WETH_TOKEN_ADDRESS} from '@yearn-finance/web-lib/utils/constants';
 import * as format from '@yearn-finance/web-lib/utils/format';
+import {formatBN} from '@yearn-finance/web-lib/utils/format';
 import performBatchedUpdates from '@yearn-finance/web-lib/utils/performBatchedUpdates';
 import * as providers from '@yearn-finance/web-lib/utils/web3/providers';
 
-import type {BigNumber} from 'ethers';
+import type {BigNumber, ethers} from 'ethers';
 import type {DependencyList} from 'react';
 import type {TBalanceData, TDefaultStatus} from '@yearn-finance/web-lib/hooks/types';
 import type {TAddress} from '@yearn-finance/web-lib/utils/address';
@@ -125,7 +125,7 @@ export function	useBalances(props?: TUseBalancesReq): TUseBalancesRes {
 				const	{token} = element;
 				const	balanceOf = results[rIndex++] as BigNumber;
 				const	decimals = results[rIndex++] as number;
-				const	rawPrice = format.BN(props?.prices?.[toAddress(token)] || ethers.constants.Zero);
+				const	rawPrice = formatBN(props?.prices?.[toAddress(token)]);
 				let symbol = results[rIndex++] as string;
 
 				if (toAddress(token) === ETH_TOKEN_ADDRESS) {
@@ -255,7 +255,7 @@ export function	useBalances(props?: TUseBalancesReq): TUseBalancesRes {
 	const assignPrices = useCallback((_rawData: TDict<TBalanceData>): TDict<TBalanceData> => {
 		for (const key of Object.keys(_rawData)) {
 			const	tokenAddress = toAddress(key);
-			const	rawPrice = format.BN(props?.prices?.[tokenAddress] || ethers.constants.Zero);
+			const	rawPrice = formatBN(props?.prices?.[tokenAddress]);
 			_rawData[tokenAddress] = {
 				..._rawData[tokenAddress],
 				rawPrice,
--- a/apps/common/utils/actions/approveToken.tsx
+++ b/apps/common/utils/actions/approveToken.tsx
@@ -1,4 +1,5 @@
 import {ethers} from 'ethers';
+import {Zero} from '@yearn-finance/web-lib/utils/format.bigNumber';
 
 export async function	isApprovedERC20(
 	provider: ethers.providers.Web3Provider,
@@ -38,7 +39,7 @@ export async function	approvedERC20Amount(
 		);
 		return await contract.allowance(address, spender);
 	} catch (error) {
-		return ethers.constants.Zero;
+		return Zero;
 	}
 }
 
--- a/apps/common/utils/constants.tsx
+++ b/apps/common/utils/constants.tsx
@@ -1,4 +1,0 @@
-import {toAddress} from '@yearn-finance/web-lib/utils/address';
-
-export const COW_VAULT_RELAYER_ADDRESS = toAddress('0xC92E8bdf79f0507f65a392b0ab4667716BFE0110');
-export const WIDO_RELAYER_ADDRESS = toAddress('0x7Fb69e8fb1525ceEc03783FFd8a317bafbDfD394');
--- a/apps/common/utils/index.tsx
+++ b/apps/common/utils/index.tsx
@@ -2,7 +2,7 @@ import {ethers} from 'ethers';
 import request from 'graphql-request';
 import {toAddress} from '@yearn-finance/web-lib/utils/address';
 import {LPYCRV_TOKEN_ADDRESS, YCRV_CURVE_POOL_ADDRESS, YVBOOST_TOKEN_ADDRESS, YVECRV_TOKEN_ADDRESS} from '@yearn-finance/web-lib/utils/constants';
-import {formatToNormalizedValue} from '@yearn-finance/web-lib/utils/format.bigNumber';
+import {formatBN, formatToNormalizedValue} from '@yearn-finance/web-lib/utils/format.bigNumber';
 import {formatAmount} from '@yearn-finance/web-lib/utils/format.number';
 
 import type {BigNumber} from 'ethers';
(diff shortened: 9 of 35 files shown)


apps/vaults/contexts/useSolver.tsx
```diff
@@ -0,0 +1,131 @@
...
+import {useSolverVanilla} from '@vaults/hooks/useSolverVanilla';
+import {useWeb3} from '@yearn-finance/web-lib/contexts/useWeb3';
+import {toAddress} from '@yearn-finance/web-lib/utils/address';
+import performBatchedUpdates from '@yearn-finance/web-lib/utils/performBatchedUpdates';
+import {DefaultTNormalizedBN} from '@common/utils';
+
+import type {TNormalizedBN} from '@common/types/types';
+import type {TWithSolver} from '@vaults/types/solvers';
+
+export enum	Solvers {
```
karelianpie, line 15:
```suggestion
export enum	Solver {
```
nit: Common pattern is to have enums singular, refer [here](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums)

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+	const [allowanceFrom, set_allowanceFrom] = useState<TNormalizedBN>();
+	const {
+		selectedOptionFrom, selectedOptionTo,
+		amount, onChangeAmount,
+		maxDepositPossible, isDepositing
+	} = useActionFlow();
+	const {onApprove, onExecuteDeposit, onExecuteWithdraw, currentSolver} = useSolver();
+	const retrieveAllowance = useAllowanceFetcher();
+
+	const shouldUseVanillaAllowance = [Solvers.VANILLA, Solvers.CHAIN_COIN, Solvers.PARTNER_CONTRACT].includes(currentSolver);
```
karelianpie, line 39: For example, move this to a 
```typescript
function shouldUseVanillaAllowance(currentSolver: Solvers): boolean {
	return [Solvers.VANILLA, Solvers.CHAIN_COIN, Solvers.PARTNER_CONTRACT].includes(currentSolver);
}
```
and have a comment on top because it's a bit unclear what is the "vanilla" allowance, wdyt?

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+	const isInputTokenEth = selectedOptionFrom?.value === ETH_TOKEN_ADDRESS;
+	const isOutputTokenEth = selectedOptionTo?.value === ETH_TOKEN_ADDRESS;
+	const canInteract = isActive && amount.raw.gt(0) && selectedOptionFrom && selectedOptionTo;
```
karelianpie, line 42: I would also move all this logic, the `canInteract` + `!isInputTokenEth && shouldUseVanillaAllowance` to a function

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+	const spender = useMemo((): TAddress => {
+		if (currentSolver === Solvers.CHAIN_COIN && isOutputTokenEth) {
+			return (toAddress(getEthZapperContract(safeChainID)));
+		} else if (currentSolver === Solvers.PARTNER_CONTRACT) {
```
karelianpie, line 47: nit: When there's a return inside the condition, I think readably is improved without the `else if`

```suggestion
		}
		
		if (currentSolver === Solvers.PARTNER_CONTRACT) {
```

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+	**********************************************************************************************/
+	const	{data: vanillAllowanceFrom, isLoading: isValidatingAllowance, mutate: mutateAllowance} = useSWR(
+		canInteract && !isInputTokenEth && shouldUseVanillaAllowance ?
+		[selectedOptionFrom, spender] as TAllowanceFetcher : null,
+		retrieveAllowance,
+		{revalidateOnFocus: false}
+	);
+
+	useEffect((): void => {
+		if (shouldUseVanillaAllowance && vanillAllowanceFrom) {
```
karelianpie, line 65: Given the logic [here](https://github.com/yearn/yearn.fi/pull/13/files#diff-fea3bd9af04552d32da82d563c7dffe401711a1b7448530346a67218e30bfbd7R58) I'm assuming that we will only have `vanillAllowanceFrom` if `shouldUseVanillaAllowance`, so could we just have `if (vanillAllowanceFrom)`?

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+		return (toAddress(selectedOptionTo?.value));
+	}, [currentSolver, isOutputTokenEth, networks, safeChainID, selectedOptionTo?.value]);
+
+	/* 🔵 - Yearn Finance **************************************************************************
+	** SWR hook to get the expected out for a given in/out pair with a specific amount. This hook is
+	** called every 10s or when amount/in or out changes. Calls the allowanceFetcher callback.
+	**********************************************************************************************/
+	const	{data: vanillAllowanceFrom, isLoading: isValidatingAllowance, mutate: mutateAllowance} = useSWR(
+		canInteract && !isInputTokenEth && shouldUseVanillaAllowance ?
+		[selectedOptionFrom, spender] as TAllowanceFetcher : null,
```
karelianpie, line 59: Is this casting required?
```suggestion
		[selectedOptionFrom, spender] : null,
```

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+	** Trigger a deposit web3 action, simply trying to deposit `amount` tokens to
+	** the selected vault.
+	**************************************************************************/
+	async function	onDepositToVault(): Promise<void> {
+		onExecuteDeposit(
+			set_txStatusDeposit,
+			async (): Promise<void> => {
+				await onSuccess();
+			}
+		);
```
karelianpie, line 108: Can't we just do this?
```suggestion
		onExecuteDeposit(set_txStatusDeposit, onSuccess);
```

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+	** Trigger a withdraw web3 action using the vault contract to take back
+	** some underlying token from this specific vault.
+	**************************************************************************/
+	async function	onWithdrawShares(): Promise<void> {
+		onExecuteWithdraw(
+			set_txStatusDeposit,
+			async (): Promise<void> => {
+				await onSuccess();
+			}
+		);
```
karelianpie, line 121: It should work with
```suggestion
		onExecuteWithdraw(set_txStatusDeposit, onSuccess);
```

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+	
+	/* 🔵 - Yearn Finance ******************************************************
+	** Wrapper to decide if we should use the partner contract or not
+	**************************************************************************/
+	function	onDepositOrWithdraw(): void {
+		if (isDepositing) {
+			onDepositToVault();
+		} else {
+			onWithdrawShares();
+		}
```
karelianpie, line 132: It would be a bit more readable IMO if
```suggestion
		isDepositing ? onDepositToVault() : onWithdrawShares();
```

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	console.warn(currentSolver);
```
karelianpie, line 135: delete?

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+	if (
+		(currentSolver === Solvers.VANILLA && (isDepositing &&( txStatusApprove.pending || amount.raw.gt(allowanceFrom?.raw || 0))))
```
karelianpie, line 137: We're doing a lot of `allowanceFrom?.raw || 0`, could we just make sure that `allowanceFrom.raw` if it didn't exist is 0 at this point, perhaps giving some default in the `useState()`
```typescript
const [allowanceFrom, set_allowanceFrom] = useState<TNormalizedBN>({raw: BigNumber.from(0), normalized: 0});
```

karelianpie: Also, the logic condition in this `if` is a bit 🤯 :) let's try and extract this to a function and potentially simplify?

apps/vaults/components/details/actions/QuickActionsFrom.tsx
```diff
@@ -0,0 +1,93 @@
...
+					<legend className={'font-number inline text-xs text-neutral-600 md:hidden'} suppressHydrationWarning>
+						{`You have ${formatAmount(selectedFromBalance.normalized)} ${selectedOptionFrom?.symbol || 'tokens'}`}
+					</legend>
+				</div>
+				{possibleOptionsFrom.length > 1 ? (
+					<Dropdown
+						defaultOption={possibleOptionsFrom[0]}
+						options={possibleOptionsFrom}
+						selected={selectedOptionFrom}
+						onSelect={(option: TDropdownOption): void => onUpdateSelectedOptionFrom(option)} />
```
karelianpie, line 44: We need any opportunity we can get to reduce the verbosity of the code to make it more readable :)
```suggestion
						onSelect={onUpdateSelectedOptionFrom} />
```

apps/vaults/components/details/actions/QuickActionsSwitch.tsx
```diff
@@ -0,0 +1,34 @@
...
+					<p>{'Deposit / Withdraw'}</p>
+				</span>
+			</div>
+			<legend className={'hidden text-xs md:inline'}>&nbsp;</legend>
+		</div>
+	);
+}
+
+
+export default VaultDetailsQuickActionsSwitch;
```
karelianpie, line 34: nit: We should not forget the new lines, multiple files missing them

apps/vaults/components/details/actions/QuickActionsTo.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import type {ReactElement} from 'react';
+import type {TDropdownOption} from '@common/types/types';
+
+function	VaultDetailsQuickActionsTo(): ReactElement {
+	const {
+		currentVault,
+		possibleOptionsTo, selectedOptionTo, onUpdateSelectedOptionTo,
+		isDepositing
+	} = useActionFlow();
```
karelianpie, line 18: nit: Looks confusing otherwise
```suggestion
	const {currentVault, possibleOptionsTo, selectedOptionTo, onUpdateSelectedOptionTo, isDepositing} = useActionFlow();
```

apps/vaults/components/details/actions/QuickActionsTo.tsx
```diff
@@ -0,0 +1,88 @@
...
+					<legend className={'font-number inline text-xs text-neutral-600 md:hidden'} suppressHydrationWarning>
+						{`APY ${formatPercent((isDepositing ? currentVault?.apy?.net_apy || 0 : 0) * 100)}`}
+					</legend>
+				</div>
+				{possibleOptionsTo.length > 1 ? (
+					<Dropdown
+						defaultOption={possibleOptionsTo[0]}
+						options={possibleOptionsTo}
+						selected={selectedOptionTo}
+						onSelect={(option: TDropdownOption): void => onUpdateSelectedOptionTo(option)} />
```
karelianpie, line 39:
```suggestion
						onSelect={onUpdateSelectedOptionTo} />
```

apps/vaults/components/details/actions/QuickActionsTo.tsx
```diff
@@ -0,0 +1,88 @@
...
+import {toAddress} from '@yearn-finance/web-lib/utils/address';
+import {formatCounterValue} from '@yearn-finance/web-lib/utils/format.value';
+import {Dropdown} from '@common/components/TokenDropdown';
+import {useTokenPrice} from '@common/hooks/useTokenPrice';
+import {formatPercent} from '@common/utils';
+
+import type {ReactElement} from 'react';
+import type {TDropdownOption} from '@common/types/types';
+
+function	VaultDetailsQuickActionsTo(): ReactElement {
```
karelianpie, line 13: This whole file looks very similar to `VaultDetailsQuickActionsFrom.tsx`, could we make a generic `VaultDetailsQuickActions` that takes a prop "from" | "to"? Or are they too diff that's making it more confusing?

apps/vaults/components/details/tabs/VaultDetailsStrategies.tsx
```diff
@@ -0,0 +1,177 @@
...
+									<p className={'font-number text-sm text-neutral-900'}>{strategy?.risk?.riskDetails?.protocolSafetyScore}</p>
+								</div>
+								<div className={'flex flex-row items-center justify-between'}>
+									<p className={'text-sm text-neutral-500'}>{'Team Knowledge Score'}</p>
+									<p className={'font-number text-sm text-neutral-900'}>{strategy?.risk?.riskDetails?.teamKnowledgeScore}</p>
+								</div>
+								<div className={'flex flex-row items-center justify-between'}>
+									<p className={'text-sm text-neutral-500'}>{'Testing Score'}</p>
+									<p className={'font-number text-sm text-neutral-900'}>{strategy?.risk?.riskDetails?.testingScore}</p>
+								</div>
```
karelianpie, line 103: Let's replace this with a tuple [label, value] and map, it will improve readability, something like
```typescript
{[
	['TVL Impact', strategy?.risk?.riskDetails?.TVLImpact],
	['Audit Score', strategy?.risk?.riskDetails?.auditScore],
	// ...
].map(([label, value]): ReactElement => (
	<div className={'flex flex-row items-center justify-between'} key={label}>
		<p className={'text-sm text-neutral-500'}>{label}</p>
		<p className={'font-number text-sm text-neutral-900'}>{value}</p>
	</div>
))}
```

apps/vaults/contexts/useActionFlow.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+	const [possibleOptionsFrom, set_possibleOptionsFrom] = useState<TDropdownOption[]>([]);
+	const [possibleZapOptionsFrom, set_possibleZapOptionsFrom] = useState<TDropdownOption[]>([]);
+	const [possibleOptionsTo, set_possibleOptionsTo] = useState<TDropdownOption[]>([]);
+	const [selectedOptionFrom, set_selectedOptionFrom] = useState<TDropdownOption | undefined>();
+	const [selectedOptionTo, set_selectedOptionTo] = useState<TDropdownOption | undefined>();
+	const [amount, set_amount] = useState<TNormalizedBN>(DefaultTNormalizedBN);
+
+	const isDepositing = useMemo((): boolean => (
+		!selectedOptionTo?.value ? true : toAddress(selectedOptionTo.value) === toAddress(currentVault.address)
```
karelianpie, line 66: Can we simplify this to
```suggestion
		!selectedOptionTo?.value || toAddress(selectedOptionTo.value) === toAddress(currentVault.address)
```

apps/vaults/contexts/useActionFlow.tsx
```diff
@@ -0,0 +1,226 @@
...
+			return ({
+				raw: vaultDepositLimit,
+				normalized: formatToNormalizedValue(vaultDepositLimit, currentVault.token.decimals)
+			});
+		} else {
+			return ({
+				raw: userBalance,
+				normalized: balances?.[toAddress(selectedOptionFrom?.value)]?.normalized || 0
+			});
+		}
```
karelianpie, line 90: I think the less `else`s the better :)
```suggestion
				}

		return ({
			raw: userBalance,
			normalized: balances?.[toAddress(selectedOptionFrom?.value)]?.normalized || 0
		});
```

apps/vaults/contexts/useActionFlow.tsx
```diff
@@ -0,0 +1,226 @@
...
+	const currentSolver = useMemo((): Solvers => {
+		const isInputTokenEth = selectedOptionFrom?.value === ETH_TOKEN_ADDRESS;
+		const isOutputTokenEth = selectedOptionTo?.value === ETH_TOKEN_ADDRESS;
+		if (isInputTokenEth || isOutputTokenEth) {
+			return Solvers.CHAIN_COIN;
+		} else if (selectedOptionFrom?.solveVia === Solvers.COWSWAP) {
+			return Solvers.COWSWAP;			
+		} else if (isDepositing && isUsingPartnerContract) {
+			return Solvers.PARTNER_CONTRACT;
+		}
```
karelianpie, line 102:
```suggestion
				if (isInputTokenEth || isOutputTokenEth) {
			return Solvers.CHAIN_COIN;
		}
		if (selectedOptionFrom?.solveVia === Solvers.COWSWAP) {
			return Solvers.COWSWAP;			
		}
		if (isDepositing && isUsingPartnerContract) {
			return Solvers.PARTNER_CONTRACT;
		}
```

apps/vaults/contexts/useActionFlow.tsx
```diff
@@ -0,0 +1,226 @@
...
+			const _possibleOptionsTo = possibleOptionsTo;
+			set_selectedOptionTo(selectedOptionFrom);
+			set_selectedOptionFrom(_selectedOptionTo);
+			set_possibleOptionsTo(possibleOptionsFrom);
+			set_possibleOptionsFrom(_possibleOptionsTo);
+			if (isDepositing) {
+				set_amount(DefaultTNormalizedBN);
+			} else {
+				set_amount(maxDepositPossible);
+			}
```
karelianpie, line 118:
```suggestion
			set_amount(isDepositing ? DefaultTNormalizedBN : maxDepositPossible);
```

apps/vaults/contexts/useSolver.tsx
```diff
@@ -0,0 +1,131 @@
...
+	const [currentSolverState, set_currentSolverState] = useState(vanilla);
+	const [isLoading, set_isLoading] = useState(false);
+
+	/* 🔵 - Yearn Finance **************************************************************************
+	** Based on the currentSolver, we initialize the solver with the required parameters.
+	**********************************************************************************************/
+	const	onUpdateSolver = useCallback(async (): Promise<void> => {
+		set_isLoading(true);
+		if (!selectedOptionFrom || !selectedOptionTo || !amount) {
+			return;
```
karelianpie, line 50: Should we `set_isLoading(false);` before returning here?

apps/vaults/hooks/useSolverCowswap.tsx
```diff
@@ -0,0 +1,300 @@
...
+	** Cowswap orders have a validity period and the return value on submit is not the execution
+	** status of the order. This method is used to check the status of the order and returns a
+	** boolean value indicating whether the order was successful or not.
+	** It will timeout once the order is no longer valid or after 50 minutes (max should be 30mn)
+	**********************************************************************************************/
+	async function checkOrderStatus(orderUID: string, validTo: Timestamp): Promise<{isSuccessful: boolean, error: Error | undefined}> {
+		for (let i = 0; i < maxIterations; i++) {
+			const {data: order} = await axios.get(`https://api.cow.fi/mainnet/api/v1/orders/${orderUID}`);
+			if (order?.status === 'fulfilled') {
+				return ({isSuccessful: true, error: undefined});
```
karelianpie, line 186: nit: Unnecessary to explicitly make it `undefined` as it's already by default
```suggestion
				return ({isSuccessful: true});
```

apps/vaults/hooks/useSolverCowswap.tsx
```diff
@@ -0,0 +1,300 @@
...
+		const	{quote, from, id} = latestQuote.current;
+		try {
+			const	buyAmountWithSlippage = getBuyAmountWithSlippage(latestQuote.current, request.current.outputToken.decimals);
+			const	{data: orderUID} = await axios.post('https://api.cow.fi/mainnet/api/v1/orders', {
+				...quote,
+				buyAmount: buyAmountWithSlippage.toString(),
+				from: from,
+				quoteId: id,
+				signature: signature.current,
+				signingScheme: String(shouldUsePresign ? 'presign' : 'eip712')
```
karelianpie, line 217: I doubt typescript would not be able to infer these types 
```suggestion
				signingScheme: shouldUsePresign ? 'presign' : 'eip712'
```

apps/vaults/hooks/useSolverCowswap.tsx
```diff
@@ -0,0 +1,300 @@
...
+	** matter the result. It returns a boolean value indicating whether the order was successful or
+	** not.
+	**********************************************************************************************/
+	const execute = useCallback(async (): Promise<boolean> => {
+		if (!latestQuote?.current || !latestQuote?.current?.quote || !request.current) {
+			return false;
+		}
+		const	{quote, from, id} = latestQuote.current;
+		try {
+			const	buyAmountWithSlippage = getBuyAmountWithSlippage(latestQuote.current, request.current.outputToken.decimals);
```
karelianpie, line 210: should this `getBuyAmountWithSlippage()` just return a `string`, it seems that's all we need, instead of calling `.toString()`

apps/vaults/components/details/actions/QuickActionsButtons.tsx
karelianpie: Nice, thanks for removing those functions. Much cleaner 👍

apps/vaults/components/details/actions/QuickActionsButtons.tsx
karelianpie: Looks better without the function here as well, nice 👍

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,134 @@
...
+			</Button>
+		);
+	}
+
+	return (
+		<Button
+			onClick={async (): Promise<void> => (
+				isDepositing ?
+					onExecuteDeposit(set_txStatusDeposit, onSuccess) :
+					onExecuteWithdraw(set_txStatusDeposit, onSuccess)
```
karelianpie, line 123: I didn't dig into this, but should this be something like `onExecuteWithdraw(set_txStatusWithdraw, onSuccess)`

apps/vaults/components/details/actions/QuickActionsButtons.tsx
karelianpie: Just a follow-up here, https://github.com/yearn/yearn.fi/pull/13#discussion_r1056110587

apps/vaults/components/details/actions/QuickActionsSwitch.tsx
karelianpie: Thanks! It probably doesn't matter too much for our case, but just for GitHub to be happy and don't show that warning is already worth it IMO

apps/vaults/components/details/actions/QuickActionsTo.tsx
karelianpie: All good, at a glance they just looked very similar. Let's keep as is, thanks!

apps/vaults/components/details/tabs/VaultDetailsStrategies.tsx
```diff
@@ -0,0 +1,176 @@
...
+	const	riskScoreElementsMap = useMemo((): TRiskScoreElementProps[] => ([
+		{label: 'TVL Impact', value: strategy?.risk?.riskDetails?.TVLImpact},
+		{label: 'Audit Score', value: strategy?.risk?.riskDetails?.auditScore},
+		{label: 'Code Review Score', value: strategy?.risk?.riskDetails?.codeReviewScore},
+		{label: 'Complexity Score', value: strategy?.risk?.riskDetails?.complexityScore},
+		{label: 'Longevity Impact', value: strategy?.risk?.riskDetails?.longevityImpact},
+		{label: 'Protocol Safety Score', value: strategy?.risk?.riskDetails?.protocolSafetyScore},
+		{label: 'Team Knowledge Score', value: strategy?.risk?.riskDetails?.teamKnowledgeScore},
+		{label: 'Testing Score', value: strategy?.risk?.riskDetails?.testingScore}
+	]), [strategy]);
```
karelianpie, line 42: nit: There's a lot of question marks and long value here, could we do something like:
```suggestion
	const	riskScoreElementsMap = useMemo((): TRiskScoreElementProps[] => {
		const {riskDetails} = strategy?.risk || {};
		return [
			{label: 'TVL Impact', value: riskDetails?.TVLImpact},
			{label: 'Audit Score', value: riskDetails?.auditScore},
			{label: 'Code Review Score', value: riskDetails?.codeReviewScore},
			{label: 'Complexity Score', value: riskDetails?.complexityScore},
			{label: 'Longevity Impact', value: riskDetails?.longevityImpact},
			{label: 'Protocol Safety Score', value: riskDetails?.protocolSafetyScore},
			{label: 'Team Knowledge Score', value: riskDetails?.teamKnowledgeScore},
			{label: 'Testing Score', value: riskDetails?.testingScore}
		];
	}, [strategy]);
```

apps/vaults/components/details/tabs/VaultDetailsStrategies.tsx
karelianpie: Nice! Even better, small follow up on this https://github.com/yearn/yearn.fi/pull/13#discussion_r1056116476

apps/vaults/contexts/useActionFlow.tsx
karelianpie: Excellent! 👍

apps/vaults/components/details/tabs/VaultDetailsStrategies.tsx
karelianpie: Also, a follow up on this, I'm not entirely sure why `riskScoreElementsMap` is using `useMemo`.

There's not much happening there :)

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+import type {ReactElement} from 'react';
+import type {SWRResponse} from 'swr';
+import type {TAddress} from '@yearn-finance/web-lib/utils/address';
+import type {TNormalizedBN} from '@common/types/types';
+
+
+function	shouldUseVanillaAllowance(solver: Solver, isInputTokenEth: boolean): boolean {
+	return ([Solver.VANILLA, Solver.CHAIN_COIN, Solver.PARTNER_CONTRACT].includes(solver) && !isInputTokenEth);
+}
```
karelianpie, line 26:
```suggestion
function	shouldUseVanillaAllowance({solver, isInputTokenEth}: {solver: Solver, isInputTokenEth: boolean}): boolean {
	return ([Solver.VANILLA, Solver.CHAIN_COIN, Solver.PARTNER_CONTRACT].includes(solver) && !isInputTokenEth);
}
```
and then
```typescript
	const withVanillaAllowance = shouldUseVanillaAllowance({
		solver: currentSolver,
		isInputTokenEth: selectedOptionFrom?.value === ETH_TOKEN_ADDRESS
	});
```

So it's a bit clearer what `selectedOptionFrom?.value === ETH_TOKEN_ADDRESS` is for

apps/vaults/components/details/tabs/VaultDetailsStrategies.tsx
karelianpie: > Also added `|| 0` just to be sure

In that case let's change the type
```
type TRiskScoreElementProps = {
	label: string;
	value?: number;
};
```
and then
```
{riskScoreElementsMap.map(({label, value}): ReactElement => (
	<RiskScoreElement
		key={label}
		label={label}
		value={value || 0} />
))}
```
so we only need one `|| 0`

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -0,0 +1,165 @@
...
+import {useWallet} from '@common/contexts/useWallet';
+import {useAllowanceFetcher} from '@common/hooks/useAllowanceFetcher';
+import {DefaultTNormalizedBN} from '@common/utils';
+
+import type {ReactElement} from 'react';
+import type {TAddress} from '@yearn-finance/web-lib/utils/address';
+import type {TAllowanceFetcher} from '@common/hooks/useAllowanceFetcher';
+import type {TNormalizedBN} from '@common/types/types';
+
+function	VaultDetailsQuickActionsButtons(): ReactElement {
```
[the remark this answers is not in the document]
karelianpie: I was able to extract UI/logic for these in the upcoming PR https://github.com/yearn/yearn.fi/pull/49/files#diff-e09560e6e81b79d7f0576a1a5d7e22f0ed2907891ecfff58302c0cb3ea66d9b2 -- so we can then plug those components here.

PR #29, opened 2023-01-03
feat: port veYFI app

xgambitox:
## Description

- Port veYFI app from [repo](https://github.com/yearn/yearn-finance-v3) into current framework and architecture.

## Motivation and Context

- Allow a better UX while navigating through all yearn apps in the same environment.

## How Has This Been Tested?

- Tested all use cases using a mainnet fork

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -40,6 +40,12 @@ module.exports = withTM(withBundleAnalyzer(withPWA({
 				has: [{type: 'host', value: 'ybribe.com'}],
 				destination: 'https://yearn.finance/ybribe/:path*',
 				permanent: true
+			},
+			{
+				source: '/:path*',
+				has: [{type: 'host', value: 'vote.yearn.finance'}],
+				destination: 'https://yearn.finance/veyfi/:path*',
+				permanent: true
 			}
 		];
 	},


apps/common/components/PageProgressBar.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+function PageProgressBar({isLoading}: TPageProgressBarProps): ReactElement {
+	const {onLoadStart, onLoadDone} = useUI();
+    
+	useClientEffect((): () => void => {
+		if (isLoading) {
+			onLoadStart();
+		} else {
+			onLoadDone();
+		}
```
karelianpie, line 18: nit: I think a bit easier to read if
```suggestion
		isLoading ? onLoadStart() : onLoadDone();
```

karelianpie: @Majorfi maybe worth to rename these in the web-lib, it feels strange that if `!isLoading` is `onLoadDone()` and when it's loading is `onLoadStart()`. Because we're missing at least a) the loading hasn't started yet; and b) the is current loading.

Also, @xgambitox I feel that this effect `PageProgressBar` would be better in the `pages/veyfi/index.tsx` as it's just returning a fragment. Could we just move the `useClientEffect` to `pages/veyfi/index.tsx` and remove this file altogether?

apps/common/components/SummaryData.tsx
```diff
@@ -0,0 +1,30 @@
...
+type TItem = {
+    label: string,
+    content: string
+}
+
+type TTabsProps = {
+	items: TItem[],
+}
+
+function SummaryData({items}: TTabsProps): ReactElement {
```
karelianpie, line 13: I would just inline the type as it's just a single prop

```suggestion
function SummaryData({items}: {items: TItem[]}): ReactElement {
```

apps/common/components/SummaryData.tsx
```diff
@@ -0,0 +1,30 @@
...
+	return (
+		<div className={'align-center flex w-full flex-row flex-wrap justify-center gap-14'}>
+			{items.map(((item): ReactElement => (
```
karelianpie, line 16: nit: I think readability would improve a bit with destructuring this
```suggestion
			{items.map((({content, label}): ReactElement => (
```

apps/veyfi/components/ClaimTab.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+function ClaimTab(): ReactElement {
+	const {provider, address, isActive, chainID} = useWeb3();
+	const {refresh: refreshBalances} = useWallet();
+	const {votingEscrow, positions, refresh: refreshVotingEscrow} = useVotingEscrow();
+	const refreshData = (): unknown => Promise.all([refreshVotingEscrow(), refreshBalances()]);
+	const [withdrawUnlocked, withdrawUnlockedStatus] = useTransaction(VotingEscrowActions.withdrawUnlocked, refreshData);
+
+	const web3Provider = provider as ethers.providers.Web3Provider;
+	const userAddress = address as TAddress;
```
karelianpie, line 25: Instead of casting here, can we have a type guard when the address is being used? Something like:
```typescript
function isTAddress(address?: string | null): address is TAddress {
	const regex = /^0x([0-9a-f][0-9a-f])*$/i;
	return !!address && regex.test(address);
}
```
and then
```typescript
const executeWithdrawUnlocked = (): void => {
	if (!votingEscrow  || !address) {
		return;
	}
	if (isTAddress(address)) {
		withdrawUnlocked(web3Provider, address, votingEscrow.address);
	}
};
```

Otherwise we might be casting a non valid address and giving it to `withdrawUnlocked`

karelianpie: @Majorfi, if we don't have it yet, I think that function could even live in the web-lib

apps/veyfi/components/ClaimTab.tsx
```diff
@@ -0,0 +1,74 @@
...
+						<p>{'Claim your YFI from expired veYFI lock.'}</p>
+					</div>
+				</div>
+			</div>
+
+			<div className={'col-span-1 grid w-full gap-6'}>
+				<div className={'grid grid-cols-1 gap-6 md:mt-14 md:grid-cols-2'}>
+					<AmountInput
+						label={'Unlocked YFI'}
+						amount={formatUnits(claimableAmount, 18)}
```
karelianpie, line 57: Are the decimals always 18?

apps/veyfi/components/LockTab.tsx
```diff
@@ -0,0 +1,174 @@
...
+import type {TMilliseconds} from '@yearn-finance/web-lib/utils/time';
+
+function LockTab(): ReactElement {
+	const [lockAmount, set_lockAmount] = useState(DefaultTNormalizedBN);
+	const [lockTime, set_lockTime] = useState('');
+	const {provider, address, isActive, chainID} = useWeb3();
+	const {refresh: refreshBalances} = useWallet();
+	const {votingEscrow, positions, allowances, isLoading: isLoadingVotingEscrow, refresh: refreshVotingEscrow} = useVotingEscrow();
+	const tokenBalance = useBalance(toAddress(votingEscrow?.token));
+	const clearLockAmount = (): unknown => set_lockAmount(DefaultTNormalizedBN);
```
karelianpie, line 32:
```suggestion
	const clearLockAmount = (): void => set_lockAmount(DefaultTNormalizedBN);
```

apps/veyfi/components/LockTab.tsx
```diff
@@ -0,0 +1,174 @@
...
+	const refreshData = (): unknown => Promise.all([refreshVotingEscrow(), refreshBalances()]);
```
karelianpie, line 33:
```suggestion
	const refreshData = async (): Promise<[void, TDict<TBalanceData>]> => Promise.all([refreshVotingEscrow(), refreshBalances()]);
```

apps/veyfi/components/LockTab.tsx
```diff
@@ -0,0 +1,174 @@
...
+	const onTxSuccess = (): unknown => Promise.all([refreshData(), clearLockAmount()]);
```
karelianpie, line 34:
```suggestion
	const onTxSuccess = async (): Promise<[[void, TDict<TBalanceData>], void]> => Promise.all([refreshData(), clearLockAmount()]);
```

apps/veyfi/components/LockTab.tsx
```diff
@@ -0,0 +1,174 @@
...
+	const [approveLock, approveLockStatus] = useTransaction(VotingEscrowActions.approveLock, refreshData);
+	const [lock, lockStatus] = useTransaction(VotingEscrowActions.lock, onTxSuccess);
+	const [increaseLockAmount, increaseLockAmountStatus] = useTransaction(VotingEscrowActions.increaseLockAmount, onTxSuccess);
+
+	const web3Provider = provider as ethers.providers.Web3Provider;
+	const userAddress = address as TAddress;
```
karelianpie, line 40: I'd suggest the same thing here as https://github.com/yearn/yearn.fi/pull/29/files#r1064538622

apps/veyfi/components/LockTab.tsx
```diff
@@ -0,0 +1,174 @@
...
+				onAction: executeIncreaseLockAmount,
+				status: increaseLockAmountStatus.loading,
+				disabled: !isActive || !isValidNetwork || !isApproved || !isValidLockAmount || !isValidLockTime || isLoadingVotingEscrow
+			}
+			: {
+				label: 'Lock',
+				onAction: executeLock,
+				status: lockStatus.loading,
+				disabled: !isActive || !isValidNetwork || !isApproved || !isValidLockAmount || !isValidLockTime || isLoadingVotingEscrow
+			};
```
karelianpie, line 119: nit: I find these nested ternary operations a bit hard to read, what you think if we refactor to something like:
```typescript
type TTxAction = {
	label: string;
	onAction: () => void;
	status?: boolean;
	disabled: boolean;
}

function getTxAction(): TTxAction {
	if (!isApproved) {
		const isDisabled = !isActive || !isValidNetwork || isApproved || isLoadingVotingEscrow;
		return {label: 'Approve', onAction: executeApprove, status: approveLockStatus.loading, disabled: isDisabled};
	}
	
	const isDisabled = !isActive || !isValidNetwork || !isApproved || !isValidLockAmount || !isValidLockTime || isLoadingVotingEscrow;

	if (hasLockedAmount) {
		return {label: 'Lock', onAction: executeIncreaseLockAmount, status: increaseLockAmountStatus.loading, disabled: isDisabled};
	}

	return {label: 'Lock', onAction: executeLock, status: lockStatus.loading, disabled: isDisabled};
}
```

apps/veyfi/components/LockTab.tsx
```diff
@@ -0,0 +1,174 @@
...
+					<p>{'Please note, governance is currently the only use for veYFI until the full platform launches ‘soon’. Stay tuned anon.'}</p>
+				</div>
+			</div>
+
+			<div className={'col-span-1 grid w-full gap-6'}>
+				<div className={'mt-0 grid grid-cols-1 gap-6 md:mt-14 md:grid-cols-2'}>
+					<AmountInput
+						label={'YFI'}
+						amount={lockAmount.normalized}
+						onAmountChange={(amount): unknown => set_lockAmount(toNormalizedBN(amount, 18))}
```
karelianpie, line 139:
```suggestion
						onAmountChange={(amount): void => set_lockAmount(toNormalizedBN(amount, 18))}
```

apps/veyfi/components/LockTab.tsx
```diff
@@ -0,0 +1,174 @@
...
+						maxAmount={tokenBalance.normalized > 0 ? tokenBalance.normalized.toFixed(18) : ''}
```
karelianpie, line 140: Same question as above, will this always be 18?

apps/veyfi/components/LockTab.tsx
karelianpie: Also, will this always be 18 decimals?

apps/veyfi/components/LockTab.tsx
```diff
@@ -0,0 +1,174 @@
...
+					<AmountInput
+						label={'Total veYFI'}
+						amount={formatUnits(votingPower, 18)}
+						disabled
+					/>
+					<Button
+						className={'w-full md:mt-7'}
+						onClick={txAction.onAction}
+						isDisabled={txAction.disabled || txAction.status}
+						isBusy={txAction.status}
```
karelianpie, line 164: This `txAction.status` sounds a bit mysterious as a boolean, so if `status` is `true` then the button is disabled? A bit confusing, also as a boolean probably should be renamed to `isStatus` or a better name :)

karelianpie: I'd actually rename `disabled` to `isDisabled` and `status` to `isLoading`

apps/veyfi/components/ManageLockTab.tsx
```diff
@@ -0,0 +1,154 @@
...
+import type {TAddress} from '@yearn-finance/web-lib/utils/address';
+
+function ManageLockTab(): ReactElement {
+	const [lockTime, set_lockTime] = useState('');
+	const {provider, address, isActive, chainID} = useWeb3();
+	const {refresh: refreshBalances} = useWallet();
+	const {votingEscrow, positions, refresh: refreshVotingEscrow} = useVotingEscrow();
+	const clearLockTime = (): unknown => set_lockTime('');
+	const refreshData = (): unknown => Promise.all([refreshVotingEscrow(), refreshBalances()]);
+	const onTxSuccess = (): unknown =>  Promise.all([refreshData(), clearLockTime()]);
```
karelianpie, line 27: As above, I'd set the expected return types here instead of `unknown`

apps/veyfi/components/ManageLockTab.tsx
```diff
@@ -0,0 +1,154 @@
...
+	const [extendLockTime, extendLockTimeStatus] = useTransaction(VotingEscrowActions.extendLockTime, onTxSuccess);
+	const [withdrawLocked, withdrawLockedStatus] = useTransaction(VotingEscrowActions.withdrawLocked, onTxSuccess);
+
+	const web3Provider = provider as ethers.providers.Web3Provider;
+	const userAddress = address as TAddress;
```
karelianpie, line 32: Same as above, it would be better to use the type guard for type safety

apps/veyfi/components/ManageLockTab.tsx
```diff
@@ -0,0 +1,154 @@
...
+					<AmountInput
+						label={'Total veYFI'}
+						amount={formatUnits(votingPower, 18)}
+						disabled
+					/>
+					<Button
+						className={'w-full md:mt-7'}
+						onClick={executeExtendLockTime}
+						isBusy={extendLockTimeStatus.loading}
+						disabled={!isActive || !isValidNetwork || !isValidLockTime || extendLockTimeStatus.loading}
```
karelianpie, line 105: Wondering if we could extract this to an external function to have all this logic in a single place, there seems to be a lot of conditions to have the buttons disabled in many places.

apps/veyfi/contexts/useVotingEscrow.tsx
```diff
@@ -0,0 +1,160 @@
...
+	allowances: TDict<BigNumber>,
+	isLoading: boolean,
+	refresh: () => void,
+}
+const defaultProps: TVotingEscrowContext = {
+	votingEscrow: undefined,
+	positions: undefined,
+	allowances: {},
+	isLoading: true,
+	refresh: (): unknown => ({})
```
karelianpie, line 56: I would probably remove the lint rule "Unexpected empty method", for these cases, so we don't need to cast the return to `unknown` in order to have defaults, /cc @Majorfi 
```typescript
const defaultProps: TVotingEscrowContext = {
	votingEscrow: undefined,
	positions: undefined,
	allowances: {},
	isLoading: true,
	// eslint-disable-next-line @typescript-eslint/no-empty-function
	refresh: (): void => {}
};
```

apps/veyfi/contexts/useVotingEscrow.tsx
```diff
@@ -0,0 +1,160 @@
...
+			decimals,
+			supply,
+			rewardPool
+		] = await ethcallProvider.tryAll([
+			veYFIContract.token(),
+			veYFIContract.name(),
+			veYFIContract.symbol(),
+			veYFIContract.decimals(),
+			veYFIContract.supply(),
+			veYFIContract.reward_pool()
```
karelianpie, line 81: As these all come as `any`, should we do some validations here before using them?

apps/veyfi/contexts/useVotingEscrow.tsx
```diff
@@ -0,0 +1,160 @@
...
+	const positionsFetcher = useCallback(async (): Promise<TVotingEscrowPosition> => {
+		if (!isActive || !address) {
+			return {};
+		}
+		const currentProvider = getProvider(1);
+		const ethcallProvider = await newEthCallProvider(currentProvider);
+		const veYFIPositionHelperContract = new Contract(VEYFI_POSITION_HELPER_ADDRESS, VEYFI_POSITION_HELPER_ABI);
+
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		const [positionDetails] = await ethcallProvider.tryAll([veYFIPositionHelperContract.getPositionDetails(address)]) as [any];
```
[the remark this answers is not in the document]
karelianpie: Ideally we'd have type guards to make sure the data matches the type, or even better, add a library like [zod](https://github.com/colinhacks/zod) to help

apps/veyfi/hooks/useTransaction.ts
```diff
@@ -0,0 +1,73 @@
...
+type TTxFunc<T1 extends TTxFuncArgs, T2> = (...args: T1) => Promise<T2>;
+
+export const useTransaction = <T extends TTxFuncArgs>(
+	func: TTxFunc<T, boolean>,
+	onSuccess?: (payload: boolean | undefined) => void,
+	onError?: (error: unknown) => void
+): [TTxFunc<T, boolean | undefined>, TStatus, boolean | undefined] => {
+	const {toast} = yToast();
+	// eslint-disable-next-line @typescript-eslint/naming-convention
+	const [result, set_result] = useState<boolean | undefined>();
```
karelianpie, line 23: How about
```suggestion
	const [isTransactionSuccessful, set_isTransactionSuccessful] = useState<boolean | undefined>();
```

apps/veyfi/hooks/useTransaction.ts
```diff
@@ -0,0 +1,73 @@
...
+		executed: isExecuted
+	};
+
+	const execute = async (...p: T): Promise<boolean | undefined> => {
+		set_isLoading(true);
+		set_error(undefined);
+		let funcResult;
+		try {
+			// eslint-disable-next-line @typescript-eslint/naming-convention
+			const funcResult = await func(...p);
```
karelianpie, line 40:
```suggestion
			const isSuccess = await func(...p);
```

apps/veyfi/utils/actions/votingEscrow.ts
```diff
@@ -0,0 +1,95 @@
...
+import type {TSeconds} from '@yearn-finance/web-lib/utils/time';
+
+const handleTx = async (txPromise: Promise<ethers.providers.TransactionResponse>): Promise<boolean> => {
+	try {
+		const tx = await txPromise;
+		const receipt = await tx.wait();
+		if (receipt.status === 0) {
+			console.error('Fail to perform transaction');
+			return false;
+		}
```
karelianpie, line 17: As we have a catch block that does exactly this, I'd just throw here
```suggestion
		if (receipt.status === 0) {
			throw new Error('Fail to perform transaction');
		}
```

apps/veyfi/utils/actions/votingEscrow.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export async function withdrawUnlocked(
+	provider: ethers.providers.Web3Provider,
+	accountAddress: TAddress,
+	votingEscrowAddress: TAddress
+): Promise<boolean> {
+	const signer = provider.getSigner(accountAddress);
+	const votingEscrowContract = new ethers.Contract(votingEscrowAddress, VEYFI_ABI, signer);
+	const {penalty} = await votingEscrowContract.callStatic.withdraw();
+	if ((penalty as BigNumber).gt(0)) {
```
karelianpie, line 79: Could we have a guard here and in other places where we expect a `BigNumber`? Feels risky to cast this to `BigNumber`

apps/veyfi/utils/actions/votingEscrow.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export async function withdrawLocked(
+	provider: ethers.providers.Web3Provider,
+	accountAddress: TAddress,
+	votingEscrowAddress: TAddress
+): Promise<boolean> {
+	const signer = provider.getSigner(accountAddress);
+	const votingEscrowContract = new ethers.Contract(votingEscrowAddress, VEYFI_ABI, signer);
+	return await handleTx(votingEscrowContract.withdraw());
```
karelianpie, line 93: I think `await` is not necessary here
```suggestion
	return handleTx(votingEscrowContract.withdraw());
```

apps/veyfi/utils/constants.ts
```diff
@@ -0,0 +1,12 @@
+import {toAddress} from '@yearn-finance/web-lib/utils/address';
+
+import type {TAddress} from '@yearn-finance/web-lib/utils/address';
+import type {TWeeks} from '@yearn-finance/web-lib/utils/time';
+
+export const VEYFI_ADDRESS: TAddress = toAddress('0x90c1f9220d90d3966FbeE24045EDd73E1d588aD5');
+export const VEYFI_POSITION_HELPER_ADDRESS: TAddress = toAddress('0x5A70cD937bA3Daec8188E937E243fFa43d6ECbe8');
+export const YFI_ADDRESS: TAddress = toAddress('0x0bc529c00C6401aEF6D220BE8C6Ea1667F6Ad93e');
+
+export const MAX_LOCK_TIME: TWeeks = 208;
```
karelianpie, line 10: nit: @Majorfi We could export a type `IntRange` in the web-lib so that we can have a bit more control on these magical numbers

```typescript
type Enumerate<N extends number, Acc extends number[] = []> = Acc['length'] extends N ? Acc[number] : Enumerate<N, [...Acc, Acc['length']]>

type IntRange<F extends number, T extends number> = Exclude<Enumerate<T>, Enumerate<F>>
```
then we could use
```typescript
type TWeeks = IntRange<1, 52> // or whatever range we want
```

apps/veyfi/utils/validations.ts
```diff
@@ -0,0 +1,81 @@
...
+  
+	return {isValid: isApproved};
+}
+
+export type TValidateAmountProps = {
+    amount: string | number;
+    balance?: string | number;
+    minAmountAllowed?: string | number;
+    maxAmountAllowed?: string | number;
+	shouldDisplayMin?: boolean;
```
karelianpie, line 35: nit: Indentation
```suggestion
    shouldDisplayMin?: boolean;
```

apps/veyfi/utils/validations.ts
```diff
@@ -0,0 +1,81 @@
...
+export function validateAmount(props: TValidateAmountProps): TValidationResponse {
+	const {amount, balance, minAmountAllowed, maxAmountAllowed, shouldDisplayMin} = props;
+	const amountNumber = Number(amount);
+	
+	if (amountNumber === 0) {
+		return {};
+	}
+  
+	if (amountNumber < 0) {
+		return {error: 'Invalid amount'};
```
karelianpie, line 47: Should we also have here and wherever the amount is not valid `isValid: false`?

apps/veyfi/utils/validations.ts
```diff
@@ -0,0 +1,81 @@
...
+export type TValidateNetworkProps = {
+	supportedNetwork: number;
+	walletNetwork?: number;
+  }
+  
+export function validateNetwork(props: TValidateNetworkProps): TValidationResponse {
+	const {supportedNetwork, walletNetwork} = props;
+  
+	if (!walletNetwork) {
+		return {error: 'Wallet Not Connected'};
```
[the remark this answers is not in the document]
karelianpie: If not, we can just default `isValid` to `false`

apps/common/utils/index.tsx
```diff
@@ -70,11 +70,11 @@ export function getAmountWithSlippage(from: string, to: string, value: BigNumber
...
 }
 
-export function handleInputChange(
-	e: React.ChangeEvent<HTMLInputElement>,
+export function toNormalizedBN(
+	value: string,
 	decimals: number
 ): TNormalizedBN {
-	let		amount = e.target.value.replace(/,/g, '.').replace(/[^0-9.]/g, '');
+	let		amount = value.replace(/,/g, '.').replace(/[^0-9.]/g, '');
```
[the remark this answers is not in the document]
karelianpie: The change is similar to this, it's using the `value` rather than the event, it's already merged so we can just reuse

https://github.com/yearn/web-lib/pull/156/files#diff-34b253f6b41c97d2870b0de3bbfc44697ea3da921a051f46ba4a4ef235adf7af

apps/veyfi/contexts/useVotingEscrow.tsx
karelianpie: Actually discussing with @Majorfi, we can do
```typescript
refresh: (): void => undefined
```
or even
```typescript
refresh: (): void => void(0)
```

karelianpie (changes requested):
Mostly concerned with the castings, otherwise looking pretty good! 👍

apps/veyfi/components/ClaimTab.tsx
karelianpie: Created a PR https://github.com/yearn/web-lib/pull/160

apps/common/components/PageProgressBar.tsx
karelianpie: Sounds good, just feels a bit strange to have a component just returning a fragment. Perhaps that component should still come from the web-lib, wdyt?

apps/veyfi/components/ClaimTab.tsx
karelianpie: Yes, that's actually even better, we can reuse the `isTAddress()` in the `useWeb3()` hook so that when you destructure the `address` is already of type TAddress. PR https://github.com/yearn/web-lib/pull/162

apps/veyfi/components/LockTab.tsx
karelianpie: Fair enough, how about the disabled state, can we DRY it a bit?

apps/veyfi/contexts/useVotingEscrow.tsx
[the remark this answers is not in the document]
karelianpie: All good, I really think we'd benefit from having [zod](https://github.com/colinhacks/zod) or a similar validation library though /cc @Majorfi

apps/veyfi/hooks/useTransaction.ts
karelianpie: Ah ok, that makes sense! 👍

apps/veyfi/utils/actions/votingEscrow.ts
karelianpie: That would be better yes. Although in this particular function there's a catch block that logs the error and returns `false`, so was suggesting we can just throw inside the `try` block so that it's caught in the next block

apps/veyfi/utils/actions/votingEscrow.ts
karelianpie: I meant more having a validation that `penalty` is indeed a `BigNumber`, something like
```typescript
if (isBigNumber(penalty) && penalty.gt(0)) {
	throw new Error('Tokens are not yet unlocked');
}
```
and
```typescript
unction isBigNumber(number: unknown): number is BigNumber {
	return BigNumber.isBigNumber(number);
}
```

karelianpie: I think it doesn't hurt to go a step extra and validate the data, but if you have trust in ethers that's fine :)

karelianpie (dismissed):

apps/veyfi/contexts/useVotingEscrow.tsx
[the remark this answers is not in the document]
karelianpie: This one https://github.com/dethcrypto/eth-sdk from the same guys looks pretty nice, only for ethereum

karelianpie (dismissed):

Issue #27, opened 2023-01-03
Feat - Hook useCurrentApp


Majorfi:
Create a basic hook to get the current App.

Ideally this should returns an identifier, but this could also returns all "specific" apps stuff (manifest, header, etc.)

        <img width="801" alt="Screenshot 2022-12-29 at 1 45 36 pm" src="https://user-images.githubusercontent.com/78794805/209946686-e3730117-ccce-4396-9aba-ea88f3e0213a.png">

```typescript
function getCurrentApp(router) {
  if (router.pathname.startsWith('/ycrv')) {
    return 'yCRV';
  }
  if (router.pathname.startsWith('/vaults')) {
    return 'Vaults';
  }
  if (router.pathname.startsWith('/ybribe')) {
    return 'yBribe';
  }
  return 'Home';
}

```

_Originally posted by @karelianpie in https://github.com/yearn/yearn.fi/pull/24#discussion_r1058905886_

karelianpie:
Strange that https://github.com/yearn/yearn.fi/pull/30#issue-1517838964 didn't close this PR

PR #30, opened 2023-01-03
feat: Implement `useCurrentApp` and `useAllApps` hooks

karelianpie:
## Description

Implements the `useCurrentApp` and `useAllApps` hooks. The reason for these two as opposed to just the `useCurrentApp` is that we have places in the app that need all menus or all headers, and it was cumbersome to have the `useCurrentApp` return the data for all them on top of for the single app route.

## Related Issue

Closes https://github.com/yearn/yearn.fi/issues/27

## Motivation and Context

DRY

## How Has This Been Tested?

Ran locally

N/A


public/apps/ybribe-manifest.json
```diff
@@ -0,0 +1,35 @@
+{
```
karelianpie, line 1: @Majorfi this is not being applied, didn't look too much into it yet. But if you have any idea then let me know

apps/common/components/Apps.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+type TApp = {
+	name: AppName;
+	href: string;
+	menu: TMenu[];
+	manifest: TMetaFile;
+	icon: ReactElement;
+}
+
+export const APPS: { [key in AppName]: TApp } = {
```
karelianpie, line 31: Noticed that we already had here a lot of static data, so made this a first class citizen as we have everything in a single place.

It allowed to DRY the code quite a bit

apps/common/components/Apps.tsx
```diff
@@ -0,0 +1,77 @@
...
+import type {ReactElement} from 'react';
+import type {TMenu} from '@yearn-finance/web-lib/layouts/Header.next';
+import type {TMetaFile} from './Meta';
+
+export enum AppName {
+	VAULTS = 'vaults',
+	YCRV = 'yCRV',
+	VEYFI = 'veYFI',
+	YBRIBE = 'yBribe'
+}
```
karelianpie, line 21: Was hoping to use this to have the name of the apps in a single place, didn't go and replace it all over but could do in this PR. Wdyt @Majorfi?

karelianpie:
> Checking a bit here! The logo animation is currently broken :(

I believe it's not related to the changes in this PR, I also do get sometime a broken animation when the server restarts or in other situations, it appears to be flaky.

apps/ybribe/Wrapper.tsx
```diff
@@ -10,22 +12,35 @@ import type {ReactElement} from 'react';
...
+							variants={variants}>
+							<div className={'md:mb-0 md:mt-16'}>
+								<div className={'mx-auto flex w-full max-w-6xl flex-col items-center justify-center'}>
+									<div className={'mt-10 w-[300px] md:w-full'}>
+										<div className={'flex w-full items-center justify-center text-center text-4xl font-bold uppercase text-neutral-900 md:text-8xl'}>
+											<HeroTimer />
+										</div>
+									</div>
+								</div>
+							</div>
```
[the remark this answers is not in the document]
karelianpie: I was going to swear that I had removed them, maybe forgot to save the file 🤦 

Fixed with https://github.com/yearn/yearn.fi/pull/30/commits/024f0bb1bbdc2f6a3f82904dbc7d299dd73c75ca

PR #56, opened 2023-01-17
feat: veYFI gauges

xgambitox:
## Description

- Add veYFI gauge use cases

## Motivation and Context

- Prepare gauge voting for veYFI

## How Has This Been Tested?

- TBD

- TBD

### DIFF
--- a/apps/veyfi/components/RedeemTab.tsx
+++ b/apps/veyfi/components/RedeemTab.tsx
@@ -104,7 +104,7 @@ function RedeemTab(): ReactElement {
 						isBusy={isLoadingOption || approveRedeemStatus.loading || redeemStatus.loading || status === 'loading'}
 						disabled={!isActive || !isValidNetwork || !isValidRedeemAmount || status === 'loading' || status === 'error'}
 					>
-						{'Redeem'}
+						{isApproved ? 'Redeem' : 'Approve'}
 					</Button>
 				</div>
 			</div>
--- a/apps/veyfi/utils/actions/option.ts
+++ b/apps/veyfi/utils/actions/option.ts
@@ -15,6 +15,6 @@ export async function redeem(
 	ethRequired: BigNumber
 ): Promise<TTxResponse> {
 	const signer = provider.getSigner(accountAddress);
-	const gaugeContract = new ethers.Contract(VEYFI_OPTIONS_ADDRESS, VEYFI_OPTIONS_ABI, signer); // TODO: update once deployed
-	return handleTx(gaugeContract.exercise(amount, accountAddress, {value: ethRequired}));
+	const optionsContract = new ethers.Contract(VEYFI_OPTIONS_ADDRESS, VEYFI_OPTIONS_ABI, signer); // TODO: update once deployed
+	return handleTx(optionsContract.exercise(amount, accountAddress, {value: ethRequired}));
 }


apps/common/components/Table.tsx
```diff
@@ -0,0 +1,97 @@
+import {useCallback, useMemo, useState} from 'react';
+import {sort} from '@veYFI/utils';
+import IconChevronPlain from '@common/icons/IconChevronPlain';
+
+import type {ReactElement} from 'react';
+
+type TSortOrder = 'asc' | 'desc' | '';
```
karelianpie, line 7: nit: I think here we could go with `null` as it's more aligned with what the order is, wdyt?

```suggestion
type TSortOrder = 'asc' | 'desc' | null;
```

apps/common/components/Table.tsx
```diff
@@ -0,0 +1,97 @@
...
+	fullWidth?: boolean;
+	columnSpan?: number;
+	format?: (item: T) => string;
+	transform?: (item: T) => ReactElement;
+}
+  
+type TTableProps<T> = {
+	metadata: TMetadata<T>[];
+	data: T[];
+	columns?: number;
```
karelianpie, line 30: Perhaps a philosophical question here, but is it a table if there're no columns?

apps/common/components/Table.tsx
```diff
@@ -0,0 +1,97 @@
...
+			{sortedData.map((item, rowIndex): ReactElement => {
+				return (
+					<div 
+						key={`row_${rowIndex}`} 
+						className={`grid w-full grid-cols-1 border-t border-neutral-200 px-4 py-2 transition-colors hover:bg-neutral-300 md:border-none md:px-8 ${`md:grid-cols-${numberOfColumns}`} ${onRowClick ? 'cursor-pointer' : 'cursor-auto'}`}
+						onClick={(): void => onRowClick?.(item)}
+					>
+						{metadata.map(({key, label, className, fullWidth, columnSpan, format, transform}): ReactElement => {
+							// eslint-disable-next-line @typescript-eslint/no-explicit-any
+							const isNumber = !isNaN(item[key] as any);
```
karelianpie, line 78: I reckon here we could do just cast `item[key]` to number, as that's what `isNaN` wants, and if it's not then it would just return `true`. So we should be safe

```suggestion
							const isNumber = !isNaN(item[key] as number);
```

apps/common/components/Table.tsx
```diff
@@ -0,0 +1,97 @@
...
+							const isZero = isNumber && Number(item[key]) === 0;
```
karelianpie, line 79: At this point, `item[key]` must be a number, so we can simply
```suggestion
							const isZero = isNumber && item[key] === 0;
```

apps/common/components/Input.tsx
```diff
@@ -0,0 +1,53 @@
...
+function Input({
+	value,
+	onChange,
+	label,
+	placeholder,
+	legend,
+	error,
+	disabled,
+	loading
+}: TAmountInputProps): ReactElement {
```
karelianpie, line 23: Just for readability, when it gets to these tower of pisa props, wdyt we just inline the props like so

```suggestion
function Input(props: TAmountInputProps): ReactElement {
	const {value, onChange, label, placeholder, legend, error, disabled, loading} = props;
```

apps/common/components/Input.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+type TAmountInputProps = {
+	value: string;
+	onChange?: (value: string) => void;
+	label?: string;
+	placeholder?: string;
+	legend?: string;
+	error?: string;
+	disabled?: boolean;
+	loading?: boolean;
```
karelianpie, line 11:
```suggestion
	isDisabled?: boolean;
	isLoading?: boolean;
```

apps/common/components/Input.tsx
```diff
@@ -0,0 +1,53 @@
...
+			)}
+			<div className={'relative flex w-full items-center justify-center'}>
+				<input
+					className={`h-10 w-full p-2 font-mono text-base font-normal outline-none ${error ? 'border border-solid border-[#EA5204] focus:border-[#EA5204]' : 'border-0 border-none'} ${disabled ? 'bg-neutral-300 text-neutral-600' : 'bg-neutral-0'}`}
+					type={'text'}
+					min={0}
+					aria-label={label}
+					value={value}
+					onChange={onChange ? (e): void => onChange(e.target.value) : undefined}
+					placeholder={loading ? '' : placeholder ?? ''}
```
karelianpie, line 40: Looks like `placeholder` can take `undefined`, is it not being handled as expected and hence the empty string?
```suggestion
					placeholder={loading ? undefined : placeholder}
```

apps/common/components/Input.tsx
```diff
@@ -0,0 +1,53 @@
...
+		<div className={'w-full'}>
+			{label && (
+				<p
+					className={'mb-1 w-full overflow-hidden text-ellipsis whitespace-nowrap text-base text-neutral-600'}>
+					{label}
+				</p>
+			)}
+			<div className={'relative flex w-full items-center justify-center'}>
+				<input
+					className={`h-10 w-full p-2 font-mono text-base font-normal outline-none ${error ? 'border border-solid border-[#EA5204] focus:border-[#EA5204]' : 'border-0 border-none'} ${disabled ? 'bg-neutral-300 text-neutral-600' : 'bg-neutral-0'}`}
```
karelianpie, line 34: Do you know if there's a better way to handle these states rather than having all these ternary ops scattered around the `className` string? Looks hard to read, if there would be a better way that would be great :)

apps/common/components/Input.tsx
```diff
@@ -0,0 +1,53 @@
...
+					onChange={onChange ? (e): void => onChange(e.target.value) : undefined}
+					placeholder={loading ? '' : placeholder ?? ''}
+					disabled={disabled}
+				/>
+			</div>
+			{(error || legend) && (
+				<legend className={`mt-1 pl-2 text-xs md:mr-0 ${error ? 'text-[#EA5204]' : 'text-neutral-600'}`} suppressHydrationWarning>
+					{error ?? legend}
+				</legend>
+			)}
```
karelianpie, line 48: Don't we run the risk here to have a falsey `error` being displayed instead of the `legend`? I'd refactor to
```suggestion
			{(error ?? legend) && (
				<legend className={`mt-1 pl-2 text-xs md:mr-0 ${error ? 'text-[#EA5204]' : 'text-neutral-600'}`} suppressHydrationWarning>
					{error ?? legend}
				</legend>
			)}
```

apps/common/components/Table.tsx
```diff
@@ -0,0 +1,97 @@
...
+  
+type TTableProps<T> = {
+	metadata: TMetadata<T>[];
+	data: T[];
+	columns?: number;
+	initialSortBy?: Extract<keyof T, string>;
+	onRowClick?: (item: T) => void;
+}
+
+function Table<T>({metadata, data, columns, initialSortBy, onRowClick}: TTableProps<T>): ReactElement {
```
karelianpie, line 35: Table looks fine 👍 I agree that these generic components should be the way forward

apps/veyfi/components/VoteTab.tsx
```diff
@@ -0,0 +1,81 @@
...
+	return (
+		<div className={'grid grid-cols-1 gap-6 md:grid-cols-2 md:gap-16'}>
+			<div className={'col-span-1 grid w-full gap-8'}>
+				<div className={'md:min-h-[104px]'}>
+					<h2 className={'m-0 text-2xl font-bold'}>
+						{'Vote for Gauge'}
+					</h2>
+					<div className={'mt-6 text-neutral-600'} >
+						<p>{'Vote to direct future YFI rewards to a particular gauge.'}</p>
+						<br />
```
karelianpie, line 42: 😲 we should add gap in the parent `div` :)

apps/veyfi/utils/index.ts
```diff
@@ -16,3 +17,39 @@ export function getVotingPower(lockAmount: BigNumber, unlockTime: TMilliseconds)
...
+export const keyBy = <T1, T2 extends keyof T1 & string>(array: T1[], key: T2): TDict<T1 | undefined> => 
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	(array || []).reduce((r, x): TDict<T1> => ({...r, [(x as any)[key]]: x}), {});
+
+export const sort = <T>(data: T[], by: Extract<keyof T, string>, order?: 'asc' | 'desc'): T[] => {
+	const compare = (a: T, b: T): number => {
+		const elementA = a[by];
+		const elementB = b[by];
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		if (!isNaN(elementA as any) && !isNaN(elementB as any)) {
```
karelianpie, line 30:
```suggestion
		if (!isNaN(elementA as number) && !isNaN(elementB as number)) {
```

apps/veyfi/utils/index.ts
```diff
@@ -16,3 +17,39 @@ export function getVotingPower(lockAmount: BigNumber, unlockTime: TMilliseconds)
 	}
 	return lockAmount.div(MAX_LOCK).mul(duration);
 }
+
+export const keyBy = <T1, T2 extends keyof T1 & string>(array: T1[], key: T2): TDict<T1 | undefined> => 
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	(array || []).reduce((r, x): TDict<T1> => ({...r, [(x as any)[key]]: x}), {});
```
karelianpie, line 23: Wouldn't `x[key]` be `T2` here? Which in turn is a `string`?
```suggestion
	(array || []).reduce((r, x): TDict<T1> => ({...r, [x[key] as T2]: x}), {});
```

apps/veyfi/utils/index.ts
karelianpie: I think even better, how about this:
```typescript
		if (isNumber(elementA) && isNumber(elementB)) {
			return order === 'desc' ? elementA - elementB : elementB - elementA;
		}
```
and then have a type guard that guarantees the value is a number, like:
```typescript
function isNumber(value: unknown): value is number {
	return typeof value === 'number';
}
```

apps/veyfi/utils/index.ts
```diff
@@ -16,3 +17,39 @@ export function getVotingPower(lockAmount: BigNumber, unlockTime: TMilliseconds)
...
+
+export const sort = <T>(data: T[], by: Extract<keyof T, string>, order?: 'asc' | 'desc'): T[] => {
+	const compare = (a: T, b: T): number => {
+		const elementA = a[by];
+		const elementB = b[by];
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		if (!isNaN(elementA as any) && !isNaN(elementB as any)) {
+			return order === 'desc' ? Number(elementA) - Number(elementB) : Number(elementB) - Number(elementA);
+		}
+		if (typeof elementA === 'string' && typeof elementB === 'string') {
```
karelianpie, line 33: We could have an identical type guard here but for the strings, such as https://github.com/yearn/yearn.fi/pull/56/files?diff=unified&w=0#r1115618194

apps/veyfi/utils/index.ts
```diff
@@ -16,3 +17,39 @@ export function getVotingPower(lockAmount: BigNumber, unlockTime: TMilliseconds)
...
+			return order === 'desc' ? elementA.toLowerCase().localeCompare(elementB.toLowerCase()) : elementB.toLowerCase().localeCompare(elementA.toLowerCase());
```
karelianpie, line 34: nit: Usually it's slightly better (less risks) to convert strings to `toUpperCase()`

apps/veyfi/utils/index.ts
```diff
@@ -16,3 +17,39 @@ export function getVotingPower(lockAmount: BigNumber, unlockTime: TMilliseconds)
...
+
+	return [...data].sort(compare);
+};
+
+export const handleTx = async (txPromise: Promise<ethers.providers.TransactionResponse>): Promise<boolean> => {
+	try {
+		const tx = await txPromise;
+		const receipt = await tx.wait();
+		if (receipt.status === 0) {
+			console.error('Fail to perform transaction');
```
karelianpie, line 47: Should we show a toast here for the user?

apps/common/components/Table.tsx
karelianpie: I'm not sure that's what we want, because we do want to have a `null`.

A column that is not sorted should just have sort order as `null`, the reason is that otherwise the little arrows on top depend on that knowledge (3 they are either active asc/desc or disabled). Basically, when the table loads 1 column could be sorted (say APY desc), but all the other columns should have sorting disabled (if sortable), if not sortable then they don't have the sorting caret.

Wdyt? Or am I missing something?

apps/common/components/Table.tsx
karelianpie: Sorry my question wasn't clear, I meant if the `columns` should be required :) (as it's currently optional)

apps/common/components/Table.tsx
karelianpie: Interesting, I don't get that error. Which version of TypeScript are you using?

apps/common/components/Table.tsx
karelianpie: I don't get that, version `4.9.4` should be smart enough. Are you using a different version?

apps/veyfi/components/VoteTab.tsx
karelianpie: We shouldn't use that, `flex-col` definitely the way to go here

apps/veyfi/utils/index.ts
karelianpie: I don't understand your refactor, with the function above we guarantee that both elements in the snippet below are numbers and we won't need to cast the strings, because we are sure they are numbers. That's the point of the type guard in my comment above, to tell TypeScript that we know those elements are numbers
```typescript
    if (isNumber(elementA) && isNumber(elementB)) {
        return order === 'desc' ? Number(elementA) - Number(elementB) : Number(elementB) - Number(elementA);
}
```

apps/veyfi/utils/index.ts
karelianpie: I still don't know why you need the ` as unknown`, thinking it must be your TypeScript version.

That's not required for me, can just cast as string directly

apps/veyfi/utils/index.ts
karelianpie: A quick googling seems to confirm that `toUpperCase()` is slightly preferable, but it's a nit

apps/common/components/Table.tsx
karelianpie: > how I see it is only two sorting orders exist, either ascending or descending

There's a 3rd one, unsorted. I believe that must have been the idea with the empty string.

> Either we have a sortedBy column active or not.

Hmmm sorry to be pain 😅 but we could sort multiple columns. Imagine the following scenario. I sort A first `asc`, then B `asc` and finally C `asc`, so the table

```
| A | B | C |
|---|---|---|
| 2 | 2 | 3 |
| 1 | 1 | 2 |
| 2 | 1 | 1 |
```
becomes
```
| A | B | C |
|---|---|---|
| 1 | 1 | 1 |
| 1 | 1 | 2 |
| 2 | 2 | 3 |
```

It's a hypothetical scenario, but maybe there will be tables where I'd like to sort by more than one column.

apps/common/components/Table.tsx
karelianpie: Ah I see, ok makes sense!

apps/veyfi/components/VoteTab.tsx
karelianpie: Is the above the right commit? Still see the `<br />` there :)

apps/veyfi/utils/index.ts
karelianpie: You're totally right, should have checked the documentation for that [isNaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN) 😅

apps/common/components/Table.tsx
karelianpie: > if we need the multiple column sort, we can implement it at the time we need, wdyt?

agreed, YAGNI principle 👍

apps/veyfi/components/VoteTab.tsx
karelianpie: Ahh this github is not letting authors know which line the reviewer clicked on. I definitely clicked on the <br /> line :)

> since its standard line break?

There's no line break there, just 2 paragraphs, if it was `<p>foo<br/>bar</p>` then it would be different

karelianpie:
👋 Gambito! Is this ready for re-review?

karelianpie:
@xgambitox we could close this, it was successfully merged with `main` in https://github.com/yearn/yearn.fi/pull/248

PR #65, opened 2023-01-30
feat: optimism staking rewards

xgambitox:
## Description

- Add staking rewards use cases on Optimism

## Motivation and Context

- Users should be able to participate in incentivized vaults on Optimism network, by staking their vault tokens to receive OP tokens as reward.

## How Has This Been Tested?

- Tested all use cases on Optimism chain

### DIFF
--- a/apps/vaults/components/RewardsTab.tsx
+++ b/apps/vaults/components/RewardsTab.tsx
@@ -17,12 +17,12 @@ import {approveERC20} from '@common/utils/actions/approveToken';
 
 import type {ReactElement} from 'react';
 import type {TAddress} from '@yearn-finance/web-lib/types';
-import type {TYearnVault} from '@common/types/yearn';
+import type {TYDaemonVault} from '@common/schemas/yDaemonVaultsSchemas';
 
 const DISPLAY_DECIMALS = 10;
 const trimAmount = (amount: string | number): string => Number(Number(amount).toFixed(DISPLAY_DECIMALS)).toString(); 
 
-function RewardsTab({currentVault}: {currentVault: TYearnVault}): ReactElement {
+function RewardsTab({currentVault}: {currentVault: TYDaemonVault}): ReactElement {
 	const {provider, address, isActive} = useWeb3();
 	const {safeChainID} = useChainID();
 	const {refresh: refreshBalances} = useWallet();
--- a/apps/vaults/components/details/VaultActionsTabsWrapper.tsx
+++ b/apps/vaults/components/details/VaultActionsTabsWrapper.tsx
@@ -19,7 +19,7 @@ import performBatchedUpdates from '@yearn-finance/web-lib/utils/performBatchedUp
 import IconChevron from '@common/icons/IconChevron';
 
 import type {ReactElement} from 'react';
-import type {TYearnVault} from '@common/types/yearn';
+import type {TYDaemonVault} from '@common/schemas/yDaemonVaultsSchemas';
 
 type TTabsOptions = {
 	value: number;
@@ -41,7 +41,7 @@ function	getCurrentTab({isDepositing, hasMigration, isRetired}: {isDepositing: b
 	return tabs.find((tab): boolean => tab.value === (isDepositing ? 0 : 1)) as TTabsOptions;
 }
 
-function	VaultActionsTabsWrapper({currentVault}: {currentVault: TYearnVault}): ReactElement {
+function	VaultActionsTabsWrapper({currentVault}: {currentVault: TYDaemonVault}): ReactElement {
 	const {onSwitchSelectedOptions, isDepositing, actionParams, currentSolver} = useActionFlow();
 	const [possibleTabs, set_possibleTabs] = useState<TTabsOptions[]>([tabs[0], tabs[1]]);
 	const {stakingRewardsByVault} = useStakingRewards();


apps/common/utils/index.tsx
```diff
@@ -83,3 +83,10 @@ export function	formatDateShort(value: number): string {
 
 	return (new Intl.DateTimeFormat([locale, 'en-US'], {year: 'numeric', month: 'short', day: '2-digit'}).format(value));
 }
+
+/* 🔵 - Yearn Finance **************************************************************************
+**	Returns an object composed of each element of an array, using one of the elements 
+**  properties as its key
+**********************************************************************************************/
+export const keyBy = <T1, T2 extends keyof T1 & string>(array: T1[], key: T2): TDict<T1 | undefined> => 
+	(array || []).reduce((r, x): TDict<T1> => ({...r, [x[key] as string]: x}), {});
```
karelianpie, line 92: Looks a bit cryptic, how about having somes tests, GPT suggested the following. Probably missing some edge cases though:

```typescript
import { keyBy } from './keyBy';

type TestObject = {
  id: string;
  name: string;
};

describe('keyBy', () => {
  it('should return an empty object when the input array is empty', () => {
    const input: TestObject[] = [];
    const result = keyBy(input, 'id');
    expect(result).toEqual({});
  });

  it('should create a dictionary keyed by the specified key', () => {
    const input: TestObject[] = [
      { id: '1', name: 'Alice' },
      { id: '2', name: 'Bob' },
      { id: '3', name: 'Charlie' },
    ];
    const expected = {
      '1': { id: '1', name: 'Alice' },
      '2': { id: '2', name: 'Bob' },
      '3': { id: '3', name: 'Charlie' },
    };
    const result = keyBy(input, 'id');
    expect(result).toEqual(expected);
  });

  it('should overwrite duplicate keys with the last occurrence in the array', () => {
    const input: TestObject[] = [
      { id: '1', name: 'Alice' },
      { id: '2', name: 'Bob' },
      { id: '2', name: 'Eve' },
    ];
    const expected = {
      '1': { id: '1', name: 'Alice' },
      '2': { id: '2', name: 'Eve' },
    };
    const result = keyBy(input, 'id');
    expect(result).toEqual(expected);
  });

  it('should return an empty object when the input array is null', () => {
    const input: TestObject[] | null = null;
    const result = keyBy(input as TestObject[], 'id');
    expect(result).toEqual({});
  });

  it('should return an empty object when the input array is undefined', () => {
    const input: TestObject[] | undefined = undefined;
    const result = keyBy(input as TestObject[], 'id');
    expect(result).toEqual({});
  });
});

```

apps/vaults/components/RewardsTab.tsx
```diff
@@ -0,0 +1,142 @@
...
+import {useAllowances} from '@common/hooks/useAllowances';
+import {useBalance} from '@common/hooks/useBalance';
+import {approveERC20} from '@common/utils/actions/approveToken';
+
+import type {ReactElement} from 'react';
+import type {TAddress} from '@yearn-finance/web-lib/types';
+import type {TYearnVault} from '@common/types/yearn';
+
+const DISPLAY_DECIMALS = 10;
+const trimAmount = (amount: string | number): string => Number(Number(amount).toFixed(DISPLAY_DECIMALS)).toString();
```
karelianpie, line 22: Move to utils? Add tests? :)

apps/vaults/components/RewardsTab.tsx
```diff
@@ -0,0 +1,142 @@
...
+					<div className={'mt-2 text-neutral-600'} >
+						<p>{'Unstake your yVault tokens and your remaining $OP rewards will be claimed automatically. Boom.'}</p>
+					</div>
+				</div>
+				<div className={'flex flex-col gap-4 md:flex-row'}>
+					<Input
+						className={'w-full md:w-[216px]'}
+						label={'You have staked'}
+						legend={formatCounterValue(stakeBalance.normalized, vaultBalance.normalizedPrice)}
+						value={`${trimAmount(stakeBalance.normalized)} ${currentVault.symbol}`}
```
karelianpie, line 125: We should probably have a util function for this, `displayAmount`?

apps/vaults/components/RewardsTab.tsx
```diff
@@ -0,0 +1,142 @@
...
+						isDisabled
+					/>
+					<Button
+						className={'w-full md:mt-7 md:w-[168px]'}
+						onClick={(): unknown => unstake(provider, userAddress, toAddress(stakingRewards?.address))}
```
karelianpie, line 130: If we don't have a `stakingRewards?.address`, should we even reach here?

apps/vaults/utils/actions/stakingRewards.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export async function stake(
+	provider: ethers.providers.JsonRpcProvider,
+	accountAddress: TAddress,
+	stakingAddress: TAddress,
+	amount: BigNumber
+): Promise<TTxResponse> {
+	const signer = provider.getSigner(accountAddress);
+	const stakingRewardsContract = new ethers.Contract(stakingAddress, STAKING_REWARDS_ABI, signer);
+	return await handleTx(stakingRewardsContract.stake(amount));
```
karelianpie, line 20: Follow up on https://github.com/yearn/yearn.fi/pull/65/files#r1167806331, as we're not using the returned value there, we could do the following:

```suggestion
	await handleTx(stakingRewardsContract.stake(amount));
	return;
```

apps/vaults/utils/actions/stakingRewards.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export async function unstake(
+	provider: ethers.providers.JsonRpcProvider,
+	accountAddress: TAddress,
+	stakingAddress: TAddress
+): Promise<TTxResponse> {
+	const signer = provider.getSigner(accountAddress);
+	const stakingRewardsContract = new ethers.Contract(stakingAddress, STAKING_REWARDS_ABI, signer);
+	return await handleTx(stakingRewardsContract.exit());
```
karelianpie, line 30: Same for this and the ones below

apps/vaults/components/RewardsTab.tsx
karelianpie: All good, one last thing, as the function is taking a string, should we add a check for when `isNaN`?

apps/vaults/components/RewardsTab.tsx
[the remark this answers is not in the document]
karelianpie: Let's go with `unknown` for now then :)

karelianpie (approved):

PR #73, opened 2023-02-09
Fix/minor bug fixes and improvements

Majorfi:
## Description
Bunch of minor fixes and refactor. See details in code comment!

### DIFF
--- a/apps/vaults/components/details/VaultActionsTabsWrapper.tsx
+++ b/apps/vaults/components/details/VaultActionsTabsWrapper.tsx
@@ -7,7 +7,7 @@ import VaultDetailsQuickActionsFrom from '@vaults/components/details/actions/Qui
 import VaultDetailsQuickActionsSwitch from '@vaults/components/details/actions/QuickActionsSwitch';
 import VaultDetailsQuickActionsTo from '@vaults/components/details/actions/QuickActionsTo';
 import SettingsPopover from '@vaults/components/SettingsPopover';
-import {Flows, useActionFlow} from '@vaults/contexts/useActionFlow';
+import {Flow, useActionFlow} from '@vaults/contexts/useActionFlow';
 import IconChevron from '@common/icons/IconChevron';
 
 import type {ReactElement} from 'react';
@@ -39,7 +39,7 @@ function	VaultActionsTabsWrapper(): ReactElement {
 		if (currentVault?.migration?.available && actionParams.isReady) {
 			set_possibleTabs([tabs[1]]);
 			set_currentTab(tabs[1]);
-			onSwitchSelectedOptions(Flows.WITHDRAW);
+			onSwitchSelectedOptions(Flow.Withdraw);
 		}
 	}, [currentVault?.migration?.available, actionParams.isReady]);
 
@@ -60,7 +60,7 @@ function	VaultActionsTabsWrapper(): ReactElement {
 								key={`desktop-${tab.value}`}
 								onClick={(): void => {
 									set_currentTab(tab);
-									onSwitchSelectedOptions(Flows.SWITCH);
+									onSwitchSelectedOptions(Flow.Switch);
 								}}>
 								<p
 									title={tab.label}
@@ -80,7 +80,7 @@ function	VaultActionsTabsWrapper(): ReactElement {
 									return;
 								}
 								set_currentTab(newTab);
-								onSwitchSelectedOptions(Flows.SWITCH);
+								onSwitchSelectedOptions(Flow.Switch);
 							}}>
 							{({open}): ReactElement => (
 								<>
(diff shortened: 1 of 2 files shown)


apps/vaults/components/details/VaultActionsTabsWrapper.tsx
```diff
@@ -56,7 +74,11 @@ function	VaultActionsTabsWrapper(): ReactElement {
 					<div className={'relative z-50'}>
 						<Listbox
 							value={currentTab.label}
-							onChange={(): void => onSwitchSelectedOptions()}>
+							onChange={(value): void => {
+								const	newTab = tabs.find((tab): boolean => tab.value === Number(value)) as TTabsOptions;
+								set_currentTab(newTab);
```
karelianpie, line 79: I think we should stop with castings, and instead address the underlying issue. For example, in this case, `set_currentTab` is not happy to accept an `undefined` so we could either handle the `undefined` value or simply do:
```suggestion
								const	newTab = tabs.find((tab): boolean => tab.value === Number(value));
								if (!newTab) {
									return;
								}
								set_currentTab(newTab);
```
`tabs.find()` might return `undefined` so we should let that be and not override typescript and tell it that we are sure that's never going to be the case if that makes sense

apps/common/contexts/useYearn.tsx
```diff
@@ -54,7 +54,7 @@ export const YearnContextApp = memo(function YearnContextApp({children}: {childr
 	const {safeChainID} = useChainID();
 	const {settings: baseAPISettings} = useSettings();
 	const {address, currentPartner} = useWeb3();
-	const [zapSlippage, set_zapSlippage] = useLocalStorage<number>('yearn.finance/zap-slippage', 0.1);
+	const [zapSlippage, set_zapSlippage] = useLocalStorage<number>('yearn.finance/zap-slippage', 0.5);
```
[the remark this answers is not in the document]
karelianpie: nit: Maybe have this magical number as a const somewhere at the top?

apps/common/hooks/useBalances.tsx
```diff
@@ -33,7 +33,6 @@ export type	TUseBalancesReq = {
 	key?: string | number,
 	tokens: TUseBalancesTokens[]
 	prices?: TDict<string>,
-	refreshEvery?: 'block' | 'second' | 'minute' | 'hour' | number,
```
karelianpie, line 36: was this not being used at all?

apps/common/hooks/useBalances.tsx
```diff
@@ -182,7 +181,7 @@ export function	useBalances(props?: TUseBalancesReq): TUseBalancesRes {
 
 		performBatchedUpdates((): void => {
 			set_nonce((n): number => n + 1);
-			set_balances((b): TNDict<TDict<TBalanceData>> => ({...b, [workerForChainID]: data.current[workerForChainID].balances}));
+			set_balances((b): TNDict<TDict<TBalanceData>> => ({...b, [workerForChainID]: {...(b[workerForChainID] || {}), ...data.current[workerForChainID].balances}}));
```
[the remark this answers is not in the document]
karelianpie: A bit hard to wrap your head around what's happening there, this increased the readability
```suggestion
				set_balances((b): TNDict<TDict<TBalanceData>> => ({
					...b,
					[web3ChainID]: {
						...(b[web3ChainID] || {}),
						...data.current[web3ChainID].balances
					}
				}));
```

apps/vaults/contexts/useActionFlow.tsx
```diff
@@ -14,14 +14,27 @@ import {useYearn} from '@common/contexts/useYearn';
 
 import externalzapOutTokenList from '../../common/utils/externalZapOutTokenList.json';
 
+import type {BigNumber} from 'ethers';
 import type {ReactNode} from 'react';
+import type {TAddress} from '@yearn-finance/web-lib/utils/address';
 import type {TDropdownOption, TNormalizedBN} from '@common/types/types';
 import type {TYearnVault} from '@common/types/yearn';
 
+export enum	Flows {
```
karelianpie, line 23: Should be singular, for enums
```suggestion
export enum	Flow {
```

apps/vaults/contexts/useActionFlow.tsx
```diff
@@ -14,14 +14,27 @@ import {useYearn} from '@common/contexts/useYearn';
...
+	DEPOSIT = 'deposit', // TODO: create this flow handler
```
karelianpie, line 24: and usually not all uppercase as the pattern
```suggestion
	Deposit = 'deposit', // TODO: create this flow handler
```

karelianpie (dismissed):

apps/common/hooks/useBalances.tsx
```diff
@@ -181,8 +180,14 @@ export function	useBalances(props?: TUseBalancesReq): TUseBalancesRes {
 		data.current[workerForChainID].nonce += 1;
 
 		performBatchedUpdates((): void => {
+			set_balances((b): TNDict<TDict<TBalanceData>> => ({
+				...b,
+				[workerForChainID]: {
+					...(b[workerForChainID] || {}),
+					...data.current[workerForChainID].balances
+				}
```
karelianpie, line 188: Wonder if we can extract this logic elsewhere, it's repeated in a bunch of places

PR #96, opened 2023-03-09
feat: add Childable

Majorfi:
## Description
Add a `<Childable />` component which can be used to avoid the inline ternary for component rendering.
This work mostly like `<Suspense />` but for conditional rendering:
- Render the child if `shouldRender` is `true`
- Otherwise render `fallback` if provider
- Otherwise render `null`

Breakdown of larger cleaning effort

### DIFF
--- a/apps/common/types/yearn.tsx
+++ b/apps/common/types/yearn.tsx
@@ -3,6 +3,7 @@ import type {TAddress, TDict} from '@yearn-finance/web-lib/types';
 export type TYearnVaultStrategy = {
 	address: TAddress,
 	name: string,
+	displayName: string,
 	description: string,
 	details: {
 		keeper: TAddress,
--- a/apps/common/utils/index.tsx
+++ b/apps/common/utils/index.tsx
@@ -1,12 +1,12 @@
 import {ethers} from 'ethers';
-import request from 'graphql-request';
+import {request} from 'graphql-request';
 import {toAddress} from '@yearn-finance/web-lib/utils/address';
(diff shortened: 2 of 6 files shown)


apps/common/components/AmountInput.tsx
```diff
@@ -54,22 +56,22 @@ function AmountInput({
 					placeholder={loading ? '' : placeholder ?? '0'}
 					disabled={disabled}
 				/>
-				{maxAmount && !disabled ? (
+				<Childable shouldRender={!!maxAmount && !disabled}>
```
karelianpie, line 59: Now that I see the use, thinking if `Renderable` wouldn't be a better name

apps/ybribe/components/bribe/GaugeListRow.tsx
```diff
@@ -59,6 +60,19 @@ function	GaugeListRow({currentGauge}: {currentGauge: TCurveGauges}): ReactElemen
...
 	const	nextRewardsForCurrentGaugeMap = Object.entries(nextRewardsForCurrentGauge || {}) || [];
 
+	function	renderDefaultValueUSDFallback(): ReactElement {
+		return (
+			<div className={'flex h-auto flex-col items-end pt-0 md:h-14'}>
+				<p className={'yearn--table-data-section-item-value'}>
+					{formatUSD(0, 5, 5)}
+				</p>
+				<p className={'font-number inline-flex items-baseline text-right text-xs text-neutral-400'}>
+					{'-'}
```
karelianpie, line 70: {'-'}

karelianpie (approved):
I like this idea, but now looking at the usage I feel that the fact that the `Childable` component wraps children is not as important as the fact that it's `Renderable`... the prop `shouldRender` goes well with that name too.

Otherwise I think the prop should be `<Childable shouldGiveBirth={true} />`

PR #109, opened 2023-03-15
refactor: Proof of concept tailwindcss logic

karelianpie:
## Description

Follow up on https://github.com/yearn/yearn.fi/pull/56#discussion_r1115591864

Util function `classNames` with the sole purpose of having more readable tailwindcss class names when there's logic involved. If approved we could refactor the codebase to use that instead, or potentially move this helper function to the web lib.

## Motivation and Context

Having inline logic with tailwindcss makes it a bit hard to read sometimes.

### DIFF
--- a/apps/common/utils/classNames.ts
+++ b/apps/common/utils/classNames.ts
@@ -1,0 +1,3 @@
+export function classNames(...classes: string[]): string {
+	return classes.filter(Boolean).join(' ');
+}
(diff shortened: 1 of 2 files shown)


karelianpie:
Closing as the util function was created in https://github.com/yearn/web-lib/pull/209

Will create another PR when the web lib is deployed with the new util fn

PR #116, opened 2023-03-20
Fix Portals Solver

mrunbanked:
Fixes the error with parsing API response and transaction execution

### DIFF
--- a/apps/vaults/hooks/useSolverPortals.ts
+++ b/apps/vaults/hooks/useSolverPortals.ts
@@ -56,7 +56,7 @@ function usePortalsQuote(): [
 
 		if (canExecuteFetch) {
 			try {
-				return axios.get(`https://api.portals.fi/v1/portal/${NETWORK.get(1)}/estimate`, {params});
+				return (await axios.get(`https://api.portals.fi/v1/portal/${NETWORK.get(1)}/estimate`, {params})).data;
 			} catch (error) {
 				const	_error = error as AxiosError<ApiError>;
 				set_err(error as Error);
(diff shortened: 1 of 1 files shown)


apps/vaults/hooks/useSolverPortals.ts
```diff
@@ -56,7 +56,7 @@ function usePortalsQuote(): [
 
 		if (canExecuteFetch) {
 			try {
-				return axios.get(`https://api.portals.fi/v1/portal/${NETWORK.get(1)}/estimate`, {params});
+				return (await axios.get(`https://api.portals.fi/v1/portal/${NETWORK.get(1)}/estimate`, {params})).data;
```
karelianpie, line 59: Let's go with the same pattern as the others
```suggestion
				const {data: estimate} = await axios.get(`https://api.portals.fi/v1/portal/${NETWORK.get(1)}/estimate`, {params});
				return estimate;
```

apps/vaults/hooks/useSolverPortals.ts
```diff
@@ -150,7 +150,7 @@ export function useSolverPortals(): TSolverContext {
 		try {
 			const {tx: {value, gasLimit, ...rest}} = portals;
 			const transaction = await signer.sendTransaction({
-				value: BigNumber.from(value.hex),
+				value: BigNumber.from(value?.hex ?? 0),
```
karelianpie, line 153: How come this is needed? From the [docs](https://docs.portals.fi/api/#tag/Portal/operation/PortalController_portal) it appears that those fields are required as a response

Screenshot 2023-03-20 at 19 34 00

mrunbanked: Yep, this is a mistake on the types due to some ethers weirdness. Will be fixed in the v2 version of the endpoints

karelianpie (approved):

PR #131, opened 2023-04-13
fix/wido native coin tx

sraver:
## Description

Added the `value` argument required to be sent on the `sendTransaction`.

## Motivation and Context

When the input asset is a native coin, the value is required to be specified so the wallet send those coins to the contract.


karelianpie (approved):
LGTM 👍

PR #136, opened 2023-04-18
Feat/hash solve

Majorfi:
## Description
It may be difficult to track change of all the solver values (amount, expectedOut, inputToken, outputToken) without re-render or missed event.
To try to avoid this, I added a request hash as follow to keep track of an update. If the hash change, the request has changed and we need to update our state.
```js
const requestHash = await hash(JSON.stringify({...request, solver, expectedOut}));
```

Moreover this should fix an issue with the expectedOut for zap being 0 when it should not

### DIFF
--- a/apps/common/components/GaugeDropdown.tsx
+++ b/apps/common/components/GaugeDropdown.tsx
@@ -12,7 +12,7 @@ function DropdownItem({option}: TDropdownGaugeItemProps): ReactElement {
 		<Combobox.Option value={option}>
 			{({active}): ReactElement =>
 				<div data-active={active} className={'yearn--dropdown-menu-item w-full hover:bg-neutral-0/40'}>
-					<div className={'h-6 w-6 rounded-full'}>
+					<div className={'h-6 w-6 flex-none rounded-full'}>
 						{option?.icon ? cloneElement(option.icon) : null}
 					</div>
 					<div className={'flex w-full flex-row items-center justify-between'}>
@@ -93,8 +93,8 @@ function Dropdown({
 						onClick={(): void => set_isOpen(!isOpen)}
 						className={'flex h-10 w-full items-center justify-between bg-neutral-0 p-2 text-base text-neutral-900 md:px-3'}>
 						<div className={'relative flex flex-row items-center'}>
-							<div key={selected?.label} className={'h-6 w-6 rounded-full'}>
-								{selected?.icon ? cloneElement(selected.icon) : <div className={'h-6 w-6 rounded-full bg-neutral-500'} />}
+							<div key={selected?.label} className={'h-6 w-6 flex-none rounded-full'}>
+								{selected?.icon ? cloneElement(selected.icon) : <div className={'h-6 w-6 flex-none rounded-full bg-neutral-500'} />}
 							</div>
 							<p className={'max-w-[90%] overflow-x-hidden text-ellipsis whitespace-nowrap pl-2 font-normal text-neutral-900 scrollbar-none md:max-w-full'}>
 								<Combobox.Input
--- a/apps/common/components/TokenDropdown.tsx
+++ b/apps/common/components/TokenDropdown.tsx
@@ -19,7 +19,7 @@ function DropdownItem({option, balanceSource}: TDropdownItemProps): ReactElement
 		<Combobox.Option value={option}>
 			{({active}): ReactElement =>
 				<div data-active={active} className={'yearn--dropdown-menu-item w-full hover:bg-neutral-0/40'}>
-					<div className={'h-6 w-6 rounded-full'}>
+					<div className={'h-6 w-6 flex-none rounded-full'}>
 						{option?.icon ? cloneElement(option.icon) : null}
 					</div>
 					<div>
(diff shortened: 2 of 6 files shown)


apps/common/utils/index.tsx
```diff
@@ -83,3 +83,12 @@ export function	formatDateShort(value: number): string {
 
 	return (new Intl.DateTimeFormat([locale, 'en-US'], {year: 'numeric', month: 'short', day: '2-digit'}).format(value));
 }
+
+
+export async function hash(message: string): Promise<string> {
```
karelianpie, line 88: Can we please add a few tests for this one?

apps/vaults/contexts/useSolver.tsx
```diff
@@ -59,7 +61,7 @@ function	WithSolverContextApp({children}: {children: React.ReactElement}): React
 	const chainCoin = useSolverChainCoin();
 	const partnerContract = useSolverPartnerContract();
 	const internalMigration = useSolverInternalMigration();
-	const [currentSolverState, set_currentSolverState] = useState(vanilla);
+	const [currentSolverState, set_currentSolverState] = useState<TSolverContext & {hash: string}>({...vanilla, hash: ''});
```
karelianpie, line 64: If the `TSolverContext` always has the need for a hash, we could just augment that type

apps/vaults/contexts/useSolver.tsx
```diff
@@ -40,6 +41,7 @@ const	DefaultWithSolverContext: TWithSolver = {
 	currentSolver: Solver.VANILLA,
 	effectiveSolver: Solver.VANILLA,
 	expectedOut: toNormalizedBN(0),
+	hash: '',
```
karelianpie, line 44: Is there a reason for empty `string` vs `undefined`?

apps/vaults/hooks/useSolverCowswap.tsx
```diff
@@ -113,20 +115,53 @@ export function useSolverCowswap(): TSolverContext {
...
 	**********************************************************************************************/
 	const init = useCallback(async (_request: TInitSolverArgs, shouldLogError?: boolean): Promise<TNormalizedBN> => {
-		if (isDisabled || !_request.inputToken.solveVia?.includes(Solver.COWSWAP)) {
+		/******************************************************************************************
+		** First we need to know which token we are selling to the zap. When we are depositing, we
+		** are selling the inputToken, when we are withdrawing, we are selling the outputToken.
+		** based on that token, different checks are required to determine if the solver can be
+		** used.
+		******************************************************************************************/
+		const sellToken = _request.isDepositing ? _request.inputToken: _request.outputToken;
```
karelianpie, line 124: nit: space for readability
```suggestion
		const sellToken = _request.isDepositing ? _request.inputToken : _request.outputToken;
```

apps/vaults/hooks/useSolverWido.tsx
```diff
@@ -90,16 +92,43 @@ export function useSolverWido(): TSolverContext {
 	** call getQuote to get the current quote for the provided request.current.
 	**********************************************************************************************/
 	const init = useCallback(async (_request: TInitSolverArgs, shouldLogError?: boolean): Promise<TNormalizedBN> => {
-		if (isSolverDisabled[Solver.WIDO] || !_request.inputToken.solveVia?.includes(Solver.WIDO)) {
+		/******************************************************************************************
```
karelianpie, line 95: As we are repeating this logic for all solvers, I'd move this to a common place and cleanup individual solvers

apps/vaults/contexts/useSolver.tsx
```diff
@@ -59,7 +62,7 @@ function	WithSolverContextApp({children}: {children: React.ReactElement}): React
 	const chainCoin = useSolverChainCoin();
 	const partnerContract = useSolverPartnerContract();
 	const internalMigration = useSolverInternalMigration();
-	const [currentSolverState, set_currentSolverState] = useState(vanilla);
+	const [currentSolverState, set_currentSolverState] = useState<TSolverContext & {hash: MaybeString}>({...vanilla, hash: undefined});
```
karelianpie, line 65:
```suggestion
	const [currentSolverState, set_currentSolverState] = useState<TSolverContext & {hash: MaybeString}>({...vanilla});
```

apps/vaults/contexts/useSolver.tsx
```diff
@@ -99,23 +101,27 @@ function	WithSolverContextApp({children}: {children: React.ReactElement}): React
 				**************************************************************/
 				if (currentSolver === Solver.WIDO && !isSolverDisabled[Solver.WIDO]) {
 					if (widoQuote.status === 'fulfilled' && widoQuote?.value.raw?.gt(0)) {
+						const requestHash = await hash(JSON.stringify({...request, solver: Solver.WIDO, expectedOut: widoQuote.value.raw.toString()}));
 						performBatchedUpdates((): void => {
-							set_currentSolverState({...wido, quote: widoQuote.value});
+							set_currentSolverState({...wido, quote: widoQuote.value, hash: requestHash});
 							set_isLoading(false);
 						});
```
karelianpie, line 108: Just a thought, but this also feels very repetitive, perhaps later we can try and extract this somehow.

apps/vaults/hooks/useSolverWido.tsx
karelianpie: Let's sleep on this, I think testing this code is a bit higher priority... so that we can then safely refactor. TBD

karelianpie (approved):

PR #139, opened 2023-04-19
feat: Popover component

karelianpie:
## Description

Popover component

## Motivation and Context

Give users an easy way to send feedback (feature requests or bug reports)

## How Has This Been Tested?

Ran locally

Screenshot 2023-04-19 at 14 15 27

### DIFF
--- a/apps/common/components/Popover.tsx
+++ b/apps/common/components/Popover.tsx
@@ -1,0 +1,57 @@
+import {Fragment, useState} from 'react';
+import {usePopper} from 'react-popper';
+import {Popover as PopoverHeadlessUI, Transition} from '@headlessui/react';
+
+import type {ReactElement} from 'react';
+
+export function Popover(): ReactElement {
+	const [referenceElement, set_referenceElement] = useState<HTMLButtonElement | null>(null);
+	const [popperElement, set_popperElement] = useState<HTMLDivElement | null>(null);
+	const {styles, attributes} = usePopper(referenceElement, popperElement, {
+		modifiers: [{name: 'offset', options: {offset: [-100, 10]}}]
+	});
+    
+	return (
+		<PopoverHeadlessUI className={'relative'}>
+			<PopoverHeadlessUI.Button
+				className={'fixed bottom-5 right-5 h-10 w-10 rounded-full bg-orange-500'}
+				ref={set_referenceElement}
+			>
+				{'FA'}
+			</PopoverHeadlessUI.Button>
+			<PopoverHeadlessUI.Overlay className={'fixed inset-0 bg-black opacity-30'} />
+			<Transition
+				as={Fragment}
+				enter={'transition ease-out duration-200'}
+				enterFrom={'opacity-0'}
+				enterTo={'opacity-100'}
+				leave={'transition ease-in duration-150'}
+				leaveFrom={'opacity-100'}
+				leaveTo={'opacity-0'}
+			>
+				<PopoverHeadlessUI.Panel
+					ref={set_popperElement}
+					style={styles.popper}
+					{...attributes.popper}
+				>
+					<div className={'overflow-hidden rounded-lg bg-white p-5'}>
+						<div className={'space-y-2'}>
+							<label htmlFor={'type'}>{'type'}</label>
+							<div>
+								<select name={'type'} id={'type'}>
+									<option value={'bug'}>{'bug'}</option>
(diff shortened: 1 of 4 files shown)


apps/common/components/Popover.tsx
```diff
@@ -12,80 +16,109 @@ type TRequestType = 'bug' | 'feature';
...
+					`\t\t\t\tFrom: [${reporter}](https://etherscan.io/address/${address})` :
+					'\t\t\t\tFrom: [wallet-not-connected]',
+				`\t\t\t\tChain: ${chainID}`,
+				`\t\t\t\tWallet: ${walletType}`,
+				`\t\t\t\tOrigin: [${router.asPath}](https://yearn.finance/${router.asPath})`
+			].join('\n'));
+			await axios.post('/api/notify', formData, {
+				headers: {'Content-Type': 'multipart/form-data'}
+			});
+			console.log(`${type} submitted!\n\n${description}`);
```
karelianpie, line 57: Forgotten?

apps/common/components/Popover.tsx
```diff
@@ -12,80 +16,109 @@ type TRequestType = 'bug' | 'feature';
...
 	});
-	const [type, set_type] = useState<TRequestType>('bug');
-	const [description, set_description] = useState<string>();
-	// eslint-disable-next-line @typescript-eslint/naming-convention
-	const {value, set} = useLocalStorageValue<boolean>('yearn.finance/feedback-popover');
-    
+
+	async function onSubmit(closeCallback: VoidFunction): Promise<void> {
+		const {body} = document;
+		if (body) {
```
karelianpie, line 31: nit: For less nesting, maybe we could negate this variable and just return

apps/common/components/Popover.tsx
```diff
@@ -0,0 +1,128 @@
...
+				backgroundColor: null,
+				allowTaint: true,
+				ignoreElements: (element): boolean => element.id === 'headlessui-portal-root'
+			});
+			const reporter = ens || lensProtocolHandle || (address ? truncateHex(toAddress(address), 4) : '');
+			const formData = new FormData();
+			const blob = await new Promise((resolve): void => {
+				canvas.toBlob((blob): void => resolve(blob));
+			});
+			formData.append('screenshot', blob as Blob);
```
karelianpie, line 42: Instead of casting, I'd do something along these lines because `blob` might be `null`
```suggestion
			const blob = await new Promise<Blob | null>((resolve): void => {
				canvas.toBlob((blob): void => resolve(blob));
			});
			if (blob) {
				formData.append('screenshot', blob);
			}
```

pages/api/notify.tsx
```diff
@@ -0,0 +1,47 @@
+import formidable from 'formidable';
```
karelianpie, line 1: This file should just have `.ts` extension as it doesn't have any JSX syntax.

pages/api/notify.tsx
```diff
@@ -0,0 +1,47 @@
...
+		const {fields, files} = await formData as any;
+		const {screenshot} = files;
+		await telegram.sendPhoto(
+			process.env.TELEGRAM_CHAT as string,
+			Input.fromLocalFile(screenshot.filepath),
+			{
+				caption: fields.messages,
+				parse_mode: 'Markdown'
+			}
+		);
```
karelianpie, line 35: 2 minor things:

* We could improve the types for the `formData`, check if the suggestion below makes sense;
* `process.env.TELEGRAM_CHAT` can be `undefined`, so better to check for that instead of casting;

```suggestion
		const formData = await new Promise<{fields: TRequest['body'], files: TRequest['body']}>(async (resolve, reject): Promise<void> => {
			form.parse(req, async (err: Error, fields: TRequest['body'], files: TRequest['body']): Promise<void> => {
				if (err) {
					reject('error');
				}
				resolve({fields, files});
			});
		});
		const {fields, files: {screenshot}} = formData;
		if (process.env.TELEGRAM_CHAT) {
			await telegram.sendPhoto(
				process.env.TELEGRAM_CHAT,
				Input.fromLocalFile(screenshot.filepath),
				{
					caption: fields.messages,
					parse_mode: 'Markdown'
				}
			);
		}
```

apps/common/components/Popover.tsx
karelianpie: Can we show an alert to the user instead?

PR #156, opened 2023-05-10
refactor: Coerce reports types

karelianpie:
Follow up on yDaemon reports' props type changes

* `duration`: `string` to `number`
* `durationPR`: `string` to `number`
* `APR`: `string` to `number`
* `timestamp`: `string` to `number`


apps/vaults/components/details/tabs/VaultDetailsStrategies.tsx
```diff
@@ -62,14 +62,20 @@ function	VaultDetailsStrategy({currentVault, strategy}: {currentVault: TYearnVau
 		{revalidateOnFocus: false}
 	) as SWRResponse;
 
-	const result = yDaemonReportsSchema.safeParse(data);
+	let latestApr;
 
-	if (!result.success) {
+	if (!data) {
```
karelianpie, line 67: This seems needed, as zod will try and validate even when we don't yet have the data.

PR #163, opened 2023-05-11
refactor: Try and simplify useSolver

karelianpie:
⚠️ Attempt to simplify the logic of `useSolver` to make it easier to understand and test


apps/vaults/contexts/useSolver.tsx
```diff
@@ -62,189 +71,114 @@ function	WithSolverContextApp({children}: {children: React.ReactElement}): React
...
-						const requestHash = await hash(JSON.stringify({...request, solver: 'NONE', expectedOut: '0'}));
-						performBatchedUpdates((): void => {
-							set_currentSolverState({...wido, quote: toNormalizedBN(0), hash: requestHash});
-							set_isLoading(false);
-						});
-					}
-				}
+				solvers[Solver.NONE] = {quote: {status: 'fulfilled', value: toNormalizedBN(0)}, ctx: vanilla};
+
+				const solverPriority = [Solver.WIDO, Solver.COWSWAP, Solver.PORTALS, Solver.NONE];
```
karelianpie, line 136: Wondering here if we can just use the `Solver.VANILLA` instead of `Solver.NONE`

Issue #166, opened 2023-05-16
Fix: Use client side validation for depositLimit


Majorfi:
Right now, if the depositLimit returned by yDaemon is 0 (can be because of misconfiguration or node issue), the deposit is just disabled on the website.
Even if this should not happen, website should still be usable.

To correct this, the depositLimit should be fetched on the client side for this function:

https://github.com/yearn/yearn.fi/blob/2fb10ffb261931a67c28ca7fbadf79e7ced64498/apps/vaults/contexts/useActionFlow.tsx#L108-L119

PR #172, opened 2023-05-17
feat: Use client side validation for `depositLimit`

karelianpie:
## Description

Fetch `depositLimit` on the client side with `useMaxDepositPossibleFetcher`

## Related Issue

Fixes https://github.com/yearn/yearn.fi/issues/166

## Motivation and Context

Right now, if the `depositLimit` returned by yDaemon is 0 (can be because of misconfiguration or node issue), the deposit is just disabled on the website. Even if this should not happen, website should still be usable.

## How Has This Been Tested?

Ran locally, also logs


karelianpie:
@Majorfi can you please re-approve? Github is not letting me merge, unsure what's going on

Issue #179, opened 2023-05-18
Bug: Negative zero


karelianpie:
### What happened 🆘

We're showing negative zero:

<img width="1273" alt="Yearn_Vaults" src="https://github.com/yearn/yearn.fi/assets/78794805/e2540b1c-5d71-4528-8e55-c0e9751c3019">

```json
{
    "earned": {
        "0x7D2382b1f8Af621229d33464340541Db362B4907": {
            "realizedGains": "0",
            "realizedGainsUSD": 0,
            "unrealizedGains": "-985900126704446",
            "unrealizedGainsUSD": -0.0017880314274915635
        }
    },
    "totalRealizedGainsUSD": 0,
    "totalUnrealizedGainsUSD": -0.0017880314274915635
}
```

PR #180, opened 2023-05-18
fix: Negative gains

karelianpie:
## Description

* Get the absolute value for gains, so we don't display negative dust gains;
* Cleanup / refactor `VaultDetailsHeader`.

## Related Issue

Fix https://github.com/yearn/yearn.fi/issues/179

## Motivation and Context

Avoid displaying 

image

## How Has This Been Tested?

Ran locally with negative dust gains, saw the absolute value


apps/vaults/components/details/VaultDetailsHeader.tsx
```diff
@@ -1,105 +1,100 @@
...
+	const YDAEMON_BASE_URI = `${baseAPISettings.yDaemonBaseURI || process.env.YDAEMON_BASE_URI}/${safeChainID}`;
+
+	const {data: earned} = useFetch<TYDaemonEarned>({
+		endpoint: (address && userAddress) ? `${YDAEMON_BASE_URI}/earned/${userAddress}` : null,
+		schema: yDaemonEarnedSchema
+	});
+
+	const normalizedVaultEarned = useMemo((): number => {
+		const {unrealizedGains} = earned?.earned?.[toAddress(address)] || {};
+		const absoluteUnrealizedGains = unrealizedGains ? Math.abs(Number(unrealizedGains)) : 0;
```
karelianpie, line 59: Using `Math.abs` to make the gains absolute

Issue #181, opened 2023-05-18
Bug: 


karelianpie:
### What happened 🆘

Ability to clear the input field, so we can empty it and then add numbers (maybe web-lib)

### To Reproduce 🔂

In any input field, try and delete the 0

<img width="548" alt="Screenshot 2023-05-18 at 21 17 32" src="https://github.com/yearn/yearn.fi/assets/78794805/5135280f-dd0d-4f27-8d12-112119e7c227">

PR #183, opened 2023-05-19
refactor: Use number input fields

karelianpie:
Follow up on https://github.com/yearn/web-lib/pull/245, related to https://github.com/yearn/yearn.fi/issues/181

## Description

The input fields that take numbers should be of type `number`

## Related Issue

Kind of related to https://github.com/yearn/yearn.fi/issues/181

## Motivation and Context

Prevent users from type non-digit characters.

## How Has This Been Tested?

Ran locally and was unable to input non-digit characters


karelianpie:
Closing as unnecessary

PR #192, opened 2023-05-22
feat: init yBal

Majorfi:
Init `yBal` UI, based on `yCRV`.

Draft until all questions and pending stuff are resolved to finish the implementation.


apps/common/components/AppHeader.tsx
```diff
@@ -125,7 +131,7 @@ export function	AppHeader(): ReactElement {
 	}, [pathname]);
 
 	const	supportedNetworks = useMemo((): number[] => {
-		if (pathname.startsWith('/ycrv') || pathname.startsWith('/veyfi') || pathname.startsWith('/ybribe')) {
+		if (pathname.startsWith('/ycrv') || pathname.startsWith('/ybal') || pathname.startsWith('/veyfi') || pathname.startsWith('/ybribe')) {
```
karelianpie, line 134: nit: It's starting to get a bit long, how about

```suggestion
	const mainnetPaths = ['/ycrv', '/ybal', '/veyfi', '/ybribe'];
	if (mainnetPaths.some(path => pathname.startsWith(path))) {
		return [1];
	}
```

apps/common/hooks/useCurrentApp.tsx
```diff
@@ -36,6 +37,11 @@ function useCurrentApp({pathname}: NextRouter): TCurrentApp {
 			return {name, manifest, menu, header: <YCrvHeader pathname={pathname} />};
 		}
 
+		if (pathname.startsWith('/ybal')) {
```
karelianpie, line 40: Same idea with this function, getting a little too big, how about
```typescript
function useCurrentApp({pathname}: NextRouter): TCurrentApp {
	return useMemo((): TCurrentApp => {
		const appMapping = {
			'/vaults': {appName: AppName.VAULTS, header: <VaultsHeader pathname={pathname} />},
			'/ybribe': {appName: AppName.YBRIBE, header: <YBribeHeader pathname={pathname} />},
			'/ycrv': {appName: AppName.YCRV, header: <YCrvHeader pathname={pathname} />},
			'/ybal': {appName: AppName.YBAL, header: <YBalHeader pathname={pathname} />},
			'/veyfi': {appName: AppName.VEYFI, header: <VeYfiHeader pathname={pathname} />},
		};

		for (const path in appMapping) {
			if (!pathname.startsWith(path)) {
				continue;
			}

			const {appName, header} = appMapping[path];
			const {name, manifest, menu} = APPS[appName];
			return {name, manifest, menu, header};
		}

		return {name: 'Home', manifest: homeManifest, menu: []};
	}, [pathname]);
}
```

apps/ybal/components/CardTransactorWrapper.tsx
```diff
@@ -0,0 +1,298 @@
...
+			}).perform();
+		}
+	}
+
+	/* 🔵 - Yearn Finance ******************************************************
+	** Set of memorized values to limit the number of re-rendering of the
+	** component.
+	**************************************************************************/
+	const	fromVaultAPY = useMemo((): string => {
+		if (toAddress(selectedOptionFrom.value) === STYBAL_TOKEN_ADDRESS) {
```
karelianpie, line 247: Wonder if these kinds of comparisons should be case insensitive, e.g.
```suggestion
		if (toAddress(selectedOptionFrom.value).toUpperCase() === STYBAL_TOKEN_ADDRESS.toUpperCase()) {
```

apps/ybal/components/CardTransactorWrapper.tsx
```diff
@@ -0,0 +1,298 @@
...
+	** BAL token require the allowance to be reset to 0 before being able to
+	** increase it. This function is called when the user wants to increase the
+	** allowance of the BAL token.
+	**************************************************************************/
+	async function	onIncreaseBalAllowance(): Promise<void> {
+		await new Transaction(provider, approveERC20, set_txStatusApprove).populate(
+			toAddress(selectedOptionFrom.value),
+			selectedOptionFrom.zapVia,
+			0
+		).perform();
```
karelianpie, line 183: I am finding these chained function calls a bit hard to read, I'd rather have it like this
```suggestion
		const tx = new Transaction(provider, approveERC20, set_txStatusApprove);
		tx.populate(toAddress(selectedOptionFrom.value), selectedOptionFrom.zapVia, 0);
		await tx.perform();
```

apps/ybal/components/CardZap.tsx
```diff
@@ -0,0 +1,262 @@
...
+		fromVaultAPY, toVaultAPY, expectedOutWithSlippage,
+		allowanceFrom, onApproveFrom, onZap, onIncreaseBalAllowance
+	} = useCardTransactor();
+
+	const	yBalPrice = useMemo((): number => (
+		formatToNormalizedValue(
+			formatBN(prices?.[YBAL_TOKEN_ADDRESS] || 0),
+			6
+		)
+	), [prices]);
```
karelianpie, line 40: Too many lines 😩
```suggestion
	const yBalPrice = useMemo((): number => (
		formatToNormalizedValue(formatBN(prices?.[YBAL_TOKEN_ADDRESS] || 0), 6)
	), [prices]);
```

apps/ybal/components/HarvestsListHead.tsx
```diff
@@ -0,0 +1,44 @@
...
+					<p className={'yearn--table-head-label'}>
+						{'Gain'}
+					</p>
+				</div>
+
+				<div className={'yearn--table-head-label-wrapper group col-span-2'} datatype={'number'}>
+					<p className={'yearn--table-head-label'}>
+						{'Value'}
+					</p>
+				</div>
```
karelianpie, line 26: I'd extract these to their own component, and pass the unique style as a prop

apps/ybal/components/HarvestsListRow.tsx
```diff
@@ -0,0 +1,75 @@
...
+import type {TYDaemonHarvests} from '@common/types/yearn';
+
+function	HarvestListRow({harvest}: {harvest: TYDaemonHarvests}): ReactElement {
+	return (
+		<div className={'yearn--table-wrapper'}>
+			<div className={'yearn--table-token-section'}>
+				<div className={'yearn--table-token-section-item'}>
+					<div className={'yearn--table-token-section-item-image'}>
+						<ImageWithFallback
+							alt={toAddress(harvest.vaultAddress) === STYBAL_TOKEN_ADDRESS ? 'st-yBal' : 'lp-yBal'}
```
karelianpie, line 20: Sounds a bit too optimistic, perhaps we should check if `toAddress(harvest.vaultAddress)` is a valid address before this

apps/ybal/components/HarvestsListRow.tsx
```diff
@@ -0,0 +1,75 @@
...
+				</div>
+			</div>
+
+			<div className={'yearn--table-data-section md:grid-cols-9'}>
+				<div className={'yearn--table-data-section-item md:col-span-1'} datatype={'number'}>
+					<p className={'yearn--table-data-section-item-label'}>{'Gain'}</p>
+					<b className={'yearn--table-data-section-item-value'}>
+						{formatAmount(formatToNormalizedValue(formatBN(harvest.profit).sub(formatBN(harvest.loss)), 18))}
+					</b>
+				</div>
```
karelianpie, line 39: These could be their own component as well, reduce duplication and make it easier to read the values

apps/ybal/components/HarvestsListRow.tsx
```diff
@@ -0,0 +1,75 @@
...
+						{toAddress(harvest.vaultAddress) === STYBAL_TOKEN_ADDRESS ? 'st-yBal' : 'lp-yBal'}
+					</p>
+				</div>
+			</div>
+
+			<div className={'yearn--table-data-section md:grid-cols-9'}>
+				<div className={'yearn--table-data-section-item md:col-span-1'} datatype={'number'}>
+					<p className={'yearn--table-data-section-item-label'}>{'Gain'}</p>
+					<b className={'yearn--table-data-section-item-value'}>
+						{formatAmount(formatToNormalizedValue(formatBN(harvest.profit).sub(formatBN(harvest.loss)), 18))}
```
karelianpie, line 37: We're formatting quite a bit here, I was thinking we could try and have zod parse the values to BigNumber (or other more preferable formats than what we get over the wire).

Then, we have a `toNormalizedAmount` in the web-lib that seems to be doing what we want here, e.g.
```typescript
export const	toNormalizedAmount = (v: BigNumberish, d?: number): string => (
	formatAmount(toNormalizedValue(v, d ?? 18), 6, 6)
);
```

It also defaults to 18 decimal places, so we can omit that too. Although, shall we double check that the we are getting 18 decimal places? I remember recently there was a bug where we assumed we were getting 18 and it was 6 for that particular case.

So with all the above, we could simply
```typescript
toNormalizedAmount(harvest.profit.sub(harvest.loss))
```

apps/ybal/components/CardTransactorWrapper.tsx
karelianpie: Are we sure? Most of the times we're simply casting to `TAddress`, even though they might not be

apps/ybal/contexts/useYBal.tsx
```diff
@@ -0,0 +1,235 @@
...
+			yCRVPoolAllowanceVault
+		] = await ethcallProvider.tryAll([
+			yBalContract.allowance(userAddress, ZAP_YEARN_VE_CRV_ADDRESS),
+			styBalContract.allowance(userAddress, ZAP_YEARN_VE_CRV_ADDRESS),
+			lpyBalContract.allowance(userAddress, ZAP_YEARN_VE_CRV_ADDRESS),
+			yveCRVContract.allowance(userAddress, ZAP_YEARN_VE_CRV_ADDRESS),
+			balContract.allowance(userAddress, ZAP_YEARN_VE_CRV_ADDRESS),
+			yvBoostContract.allowance(userAddress, ZAP_YEARN_VE_CRV_ADDRESS),
+			yveCRVContract.allowance(userAddress, YVECRV_POOL_LP_ADDRESS),
+			balContract.allowance(userAddress, YVECRV_POOL_LP_ADDRESS),
```
karelianpie, line 168: Is this one of the things missing? I think it would be good to add a few TODOs on these lines so we have an easier time adding what's missing

apps/ybal/components/CardTransactorWrapper.tsx
karelianpie: Perfect!

apps/ybal/components/HarvestsListRow.tsx
karelianpie: Looks good!

apps/ybal/utils/actions.ts
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		// Apply Buffer
+		const slippage = 1; // (1%)
+		const buffer = await contract.mint_buffer() as BigNumber;
+		const amountInBuffered = amountIn.mul(buffer).div(10_000);
+		const bufferedAmount = expectedAmountSwap.sub(amountInBuffered);
+
+		if (bufferedAmount.gt(expectedAmountMint)) {
+			const minOutStr = Number(ethers.utils.formatUnits(bufferedAmount, 18));
```
karelianpie, line 63: A bit misleading this variable name, as it's a `number`

apps/ybal/utils/actions.ts
```diff
@@ -0,0 +1,109 @@
...
+import type {BigNumber} from 'ethers';
+import type {TAddress} from '@yearn-finance/web-lib/types';
+import type {TTxResponse} from '@yearn-finance/web-lib/utils/web3/transaction';
+
+const ZAP_ABI = [{'inputs':[{'name':'_input_token', 'type':'address'}, {'name':'_output_token', 'type':'address'}, {'name':'_amount_in', 'type':'uint256'}, {'name':'_min_out', 'type':'uint256'}, {'name':'_recipient', 'type':'address'}, {'name':'_mint', 'type':'bool'}], 'name':'zap', 'outputs':[{'name':'', 'type':'uint256'}], 'stateMutability':'nonpayable', 'type':'function'}, {'inputs':[], 'name':'mint_buffer', 'outputs':[{'name':'', 'type':'uint256'}], 'stateMutability':'view', 'type':'function'}];
+
+const LOCAL_ZAP_YEARN_YBAL_ADDRESS = toAddress('0x5E5713a0d915701F464DEbb66015adD62B2e6AE9');
+const OUTPUT_TOKENS = [YBAL_TOKEN_ADDRESS, STYBAL_TOKEN_ADDRESS, LPYBAL_TOKEN_ADDRESS];
+
+export async function	simulateZapForMinOut(
```
karelianpie, line 16: It would be great to have this logic test, added a POC for the smaller `zap` function in this very file; https://github.com/yearn/yearn.fi/pull/192/commits/377c031acbbd1e31e0b9e151618f33808b0df908

I can try and have a go at this, wdyt?

PR #206, opened 2023-05-30
feat: migrate vaults hooks

Majorfi:
Migrate from ethers/ethcall/web3-react to wagmi/viem

2. PR to update the solvers and some actions + to prepare some next steps
3. Introduce the use of assert

**Notes:**
- Cowswap will not work as this require an upgrade based on the wagmi package they have
- Some `BigNumber` have been changed in `bigint` but just in the import, which would just not work. Fix for that will come in a PR on this PR to avoid overcomplex stuff

DO NOT MERGE

### DIFF
--- a/sentry.client.config.ts
+++ b/sentry.client.config.ts
@@ -8,7 +8,7 @@ Sentry.init({
 	dsn: process.env.SENTRY_DSN,
 
 	// Adjust this value in production, or use tracesSampler for greater control
-	tracesSampleRate: 1,
+	tracesSampleRate: 0,
 
 	// Setting this option to true will print useful information to the console while you're setting up Sentry.
 	debug: false,
--- a/sentry.edge.config.ts
+++ b/sentry.edge.config.ts
@@ -9,7 +9,7 @@ Sentry.init({
 	dsn: process.env.SENTRY_DSN,
 
 	// Adjust this value in production, or use tracesSampler for greater control
-	tracesSampleRate: 1,
+	tracesSampleRate: 0,
 
 	// Setting this option to true will print useful information to the console while you're setting up Sentry.
 	debug: false
--- a/sentry.server.config.ts
+++ b/sentry.server.config.ts
@@ -8,7 +8,7 @@ Sentry.init({
 	dsn: process.env.SENTRY_DSN,
 
 	// Adjust this value in production, or use tracesSampler for greater control
-	tracesSampleRate: 1,
+	tracesSampleRate: 0,
 
 	// Setting this option to true will print useful information to the console while you're setting up Sentry.
 	debug: false


apps/common/utils/toWagmiProvider.tsx
```diff
@@ -23,18 +24,14 @@ export async function toWagmiProvider(connector: Connector): Promise<TWagmiProvi
...
-		return ({isSuccessful: false, error: errorAsBaseError || ''});
-	}
+export type TWriteTransaction = {
+	connector: Connector;
+	contractAddress: TAddressWagmi;
+	statusHandler?: (status: typeof defaultTxStatus) => void;
+}
 
+export function assertAddress(addr: string): asserts addr is TAddress {
+	assert(isTAddress(addr), 'Address is not an address');
```
karelianpie, line 34: nit: Sounds a bit strange "Address is not an address" :)

```suggestion
	assert(isTAddress(addr), 'Address provided is invalid');
```

apps/common/utils/toWagmiProvider.tsx
```diff
@@ -23,18 +24,14 @@ export async function toWagmiProvider(connector: Connector): Promise<TWagmiProvi
...
+	assert(toAddress(addr) !== ZERO_ADDRESS, 'Address is address 0x0');
+	assert(toAddress(addr) !== ETH_TOKEN_ADDRESS, 'Address is address 0xE');
```
karelianpie, line 36:
```suggestion
	assert(toAddress(addr) !== ZERO_ADDRESS, 'Address is 0x0');
	assert(toAddress(addr) !== ETH_TOKEN_ADDRESS, 'Address is 0xE');
```

apps/vaults/components/details/actions/QuickActionsButtons.tsx
```diff
@@ -95,7 +95,7 @@ function	VaultDetailsQuickActionsButtons(): ReactElement {
 		toAddress(actionParams?.selectedOptionFrom?.value) === ETH_TOKEN_ADDRESS
 		&& toAddress(actionParams?.selectedOptionTo?.value) === YVWETH_ADDRESS
 	);
-	const hasAllowanceSet = actionParams?.amount.raw.gt(formatBN(allowanceFrom?.raw));
+	const hasAllowanceSet = actionParams?.amount.raw.gt(toBigInt(allowanceFrom?.raw));
```
karelianpie, line 98: It would be great to get rid of these parsing functions and just have `allowanceFrom?.raw` a BigInt from the get-go, perhaps coercing with Zod. It would clear the codebase quite a bit

apps/vaults/contexts/useSolver.tsx
```diff
@@ -75,7 +74,7 @@ function WithSolverContextApp({children}: { children: React.ReactElement }): Rea
 	const partnerContract = useSolverPartnerContract();
 	const internalMigration = useSolverInternalMigration();
 	const optimismBooster = useSolverOptimismBooster();
-	const [currentSolverState, set_currentSolverState] = useState<TSolverContext & { hash: MaybeString }>({...vanilla, hash: undefined});
+	const [currentSolverState, set_currentSolverState] = useState<TSolverContext & { hash?: string }>({...vanilla, hash: undefined});
```
karelianpie, line 77: nit: Since `hash` will be `undefined` if not defined, we can just do
```suggestion
	const [currentSolverState, set_currentSolverState] = useState<TSolverContext & { hash?: string }>(vanilla);
```

apps/vaults/contexts/useWalletForExternalMigrations.tsx
```diff
@@ -49,14 +47,13 @@ export const WalletForExternalMigrationsApp = memo(function WalletForExternalMig
...
-	const	{data: balances, update: updateBalances, isLoading, nonce} = useBalances({
+	const {data: balances, update: updateBalances, isLoading, nonce} = useBalances({
 		key: chainID,
-		provider: provider || getProvider(1),
 		tokens: availableTokens,
 		prices
 	});
 
-	const	onRefresh = useCallback(async (): Promise<TDict<TBalanceData>> => {
+	const onRefresh = useCallback(async (): Promise<TDict<TBalanceData>> => {
```
karelianpie, line 56: We should fix this across the whole codebase :) it's a nice idea to align everything but hard to maintain, so one space is definitely the way to do 👍

apps/vaults/hooks/useSolverChainCoin.ts
```diff
@@ -169,15 +181,24 @@ export function useSolverChainCoin(): TSolverContext {
...
 	): Promise<void> => {
-		if (!request?.current?.inputToken || !request?.current?.outputToken || !request?.current?.inputAmount) {
-			return;
+		assert(provider, 'Provider is not set');
+		assert(request.current, 'Request is not set');
+
+		//Asserting the basic & contextual request parameters
+		const {inputAmount} = request.current;
+		assert(inputAmount, 'Input amount is not set');
+		assert(inputAmount > 0n, 'Input amount is 0');
```
karelianpie, line 190: Do we need both of these or the latter would be enough?

apps/vaults/hooks/useSolverCowswap.ts
```diff
@@ -245,13 +266,13 @@ export function useSolverCowswap(): TSolverContext {
 				const {isSuccessful, error} = await checkOrderStatus(orderUID, quote.validTo);
 				if (error) {
 					console.error(error);
-					return ({isSuccessful: false, error});
+					return ({isSuccessful: false, error: error as BaseError || ''});
```
karelianpie, line 269: This empty string doesn't feel right, also worth to send these errors to Sentry.

apps/vaults/hooks/useSolverCowswap.ts
```diff
@@ -276,22 +302,35 @@ export function useSolverCowswap(): TSolverContext {
 			return (toNormalizedBN(0));
 		}
 		const quoteResult = await getQuote(_request, true);
-		return toNormalizedBN(formatBN(quoteResult?.quote?.buyAmount), _request.outputToken.decimals);
+		return (
+			toNormalizedBN(
+				toBigInt(quoteResult?.quote?.buyAmount.toString()),
```
karelianpie, line 307: Should `toBigInt` accept `BigNumberish`? So we don't need to convert to `string`?

apps/vaults/hooks/useSolverCowswap.ts
```diff
@@ -276,22 +302,35 @@ export function useSolverCowswap(): TSolverContext {
...
+			safeChainID,
+			toAddress(request.current.inputToken.value),
+			toAddress(request.current.outputToken.value),
+			toAddress(request.current.from)
+		);
 		if (existingAllowances.current[key] && !shouldForceRefetch) {
 			return existingAllowances.current[key];
 		}
+		assert(provider, 'Provider is not defined');
 		const allowance = await approvedERC20Amount(
```
karelianpie, line 334: This combo, checking if `provider` is defined before a `approvedERC20Amount` is used in a bunch of places, maybe better to move inside that function to reduce the number of asserts. Wdyt?

apps/vaults/hooks/useVaultEstimateOutFetcher.ts
```diff
@@ -0,0 +1,50 @@
...
+	const {chainID} = useWeb3();
+
+	const retrieveExpectedOut = useCallback(async (args: TVaultEstimateOutFetcher): Promise<TNormalizedBN> => {
+		const [inputToken, outputToken, inputAmount, isDepositing] = args;
+		if (isZeroAddress(inputToken?.value) || isZeroAddress(outputToken?.value) || inputAmount === 0n) {
+			return (toNormalizedBN(0));
+		}
+
+		try {
+			const decimals = toBigInt(outputToken?.decimals || 18);
```
karelianpie, line 27: Should decimals be a normal `number` rather than `bigint`?

apps/vaults/hooks/useSolverChainCoin.ts
```diff
@@ -101,44 +98,50 @@ export function useSolverChainCoin(): TSolverContext {
...
+		const result = await approveERC20({
+			connector: provider,
+			contractAddress: toWagmiAddress(request.current.inputToken.value),
+			spenderAddress: toWagmiAddress(getEthZapperContract(safeChainID)),
+			amount: amount,
+			statusHandler: txStatusSetter
+		});
+		if (result.isSuccessful) {
+			onSuccess();
 		}
```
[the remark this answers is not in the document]
karelianpie: It should go in this PR I think, since it's related

PR #209, opened 2023-05-30
lint: all \t to spaces for const, let and functions

Majorfi:
Replace:
`const\t` -> `const `
`function\t` -> `function `
`let\t` -> `let `


karelianpie (approved):
🤩

PR #219, opened 2023-06-01
Feat/wagmi address

Majorfi:
- Remove the specific WagmiAddress type to only use TAddress
- Update `toAddress` function to do legacy `toAddress` (checksum) + converting to wagmi Address type to make it usable in wagmi and ensure it's checksummed

### DIFF
--- a/apps/common/utils/toWagmiProvider.tsx
+++ b/apps/common/utils/toWagmiProvider.tsx
@@ -1,11 +1,15 @@
+import {captureException} from '@sentry/nextjs';
+import {prepareWriteContract, waitForTransaction, writeContract} from '@wagmi/core';
 import {toAddress} from '@yearn-finance/web-lib/utils/address';
 import {ETH_TOKEN_ADDRESS, ZERO_ADDRESS} from '@yearn-finance/web-lib/utils/constants';
 import {isTAddress} from '@yearn-finance/web-lib/utils/isTAddress';
+import {defaultTxStatus} from '@yearn-finance/web-lib/utils/web3/transaction';
 import {assert} from '@common/utils/assert';
 
+import type {BaseError} from 'viem';
 import type {Connector} from 'wagmi';
 import type {TAddress} from '@yearn-finance/web-lib/types';
-import type {defaultTxStatus} from '@yearn-finance/web-lib/utils/web3/transaction';
+import type {TTxResponse} from '@yearn-finance/web-lib/utils/web3/transaction';
 import type {GetWalletClientResult} from '@wagmi/core';
 
 export type TWagmiProviderContract = {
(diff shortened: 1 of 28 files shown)


apps/common/utils/toWagmiProvider.tsx
```diff
@@ -1,24 +1,24 @@
...
-	address: TAddressWagmi,
+	address: TAddress,
 }
 export async function toWagmiProvider(connector: Connector | undefined): Promise<TWagmiProviderContract> {
 	assert(connector, 'Connector is not set');
 
 	const signer = await connector.getWalletClient();
 	const chainId = await connector.getChainId();
-	const address = toWagmiAddress(signer.account.address);
+	const {address} = signer.account;
```
karelianpie, line 21: Are we certain that `address` is always there and that `signer` always has an `account`?

apps/vaults/contexts/useStakingRewards.tsx
```diff
@@ -130,7 +130,7 @@ export const StakingRewardsContextApp = memo(function StakingRewardsContextApp({
 		const calls = [];
 		for (const {address} of stakingRewards) {
 			const baseContract = {
-				address: toWagmiAddress(address),
+				address: address,
```
karelianpie, line 133: Whenever we have the same name I'd go for this for making it easier to read

```suggestion
				address,
```

apps/vaults/hooks/useSolverPartnerContract.ts
```diff
@@ -85,13 +86,16 @@ export function useSolverPartnerContract(): TSolverContext {
 		txStatusSetter: React.Dispatch<React.SetStateAction<TTxStatus>>,
 		onSuccess: () => Promise<void>
 	): Promise<void> => {
+		const partnerContract = networks[safeChainID]?.partnerContractAddress;
+
 		assert(request.current, 'Request is not set');
 		assert(request.current?.inputToken, 'Input token is not set');
+		assertAddress(partnerContract, 'partnerContract');
```
karelianpie, line 93: I think we can leave it, but it would have been better having named params `assertAddress({ thingy: partnerContract, name: 'partnerContract'})`

apps/vaults/utils/actions.ts
```diff
@@ -25,12 +25,12 @@ import type {TWriteTransaction} from '@common/utils/toWagmiProvider';
 ** @param amount - The amount of the underlying asset to deposit.
 ******************************************************************************/
 type TDepositAndStake = TWriteTransaction & {
-	vaultAddress: TAddressWagmi;
+	vaultAddress: TAddress | undefined;
```
karelianpie, line 28: nit: Can be like this for now, but usually we want to never explicitly assign anything to `undefined` so if optional I'd just do 
```suggestion
	vaultAddress?: TAddress;
```
and, if we want it to be an empty value then
```suggestion
	vaultAddress: TAddress | null;
```

PR #222, opened 2023-06-02
feat/Refactor tx handlers

Majorfi:
Create an `handleTx` function to avoid copy pasting wait/try/catch logic for every action

### DIFF
--- a/apps/vaults/contexts/useSolver.tsx
+++ b/apps/vaults/contexts/useSolver.tsx
@@ -89,7 +89,7 @@ function WithSolverContextApp({children}: { children: React.ReactElement }): Rea
 		if (currentNonce !== executionNonce.current) {
 			return;
 		}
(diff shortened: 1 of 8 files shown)


apps/common/utils/toWagmiProvider.tsx
```diff
@@ -38,3 +42,44 @@ export function assertAddress(addr: string | TAddress | undefined, name?: string
...
+type TPrepareWriteContractProp = Parameters<typeof prepareWriteContract>[0];
+
+export async function handleTx(
+	args: TWriteTransaction,
+	props: TPrepareWriteContractProp
+): Promise<TTxResponse> {
+	args.statusHandler?.({...defaultTxStatus, pending: true});
+	const wagmiProvider = await toWagmiProvider(args.connector);
+
+	//Some extra assertions
```
karelianpie, line 55: Let's remove this comment :)

PR #230, opened 2023-06-09
feat: Implement new `fetch` with schema for portals API

karelianpie:
Create new vanilla `fetch` so we don't need to change everything to hooks;

Use the new vanilla `fetch` for portals API so that we can not only get alerts in case the schema changes unexpectedly, but also have up-to-date types with a single source of truth.

### DIFF
--- a/apps/vaults/hooks/usePortalsApi.ts
+++ b/apps/vaults/hooks/usePortalsApi.ts
@@ -31,12 +31,12 @@ type TGetTransactionProps = Omit<TGetEstimateProps, 'params'> & {
 
 const portalsTxSchema = z.object({
 	to: z.string(),
-	from: z.string(),
+	from: z.string().optional(),
 	data: z.string(),
 	value: z.object({
 		type: z.string(),
 		hex: z.string()
-	}),
+	}).optional(),
 	gasLimit: z.object({
 		type: z.string(),
 		hex: z.string()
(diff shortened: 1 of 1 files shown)


apps/vaults/hooks/useSolverPortals.ts
```diff
@@ -347,3 +354,26 @@ export function useSolverPortals(): TSolverContext {
 	}), [expectedOut, init, onApprove, onExecute, onRetrieveAllowance]);
 }
 
+function isValidErrorObject(error: TPortalsError | unknown): error is TPortalsError {
```
karelianpie, line 357: @Majorfi I've created this type guard, as I did not find an error type for an object with these props `error.response.data.message`

What kind of error is this?

apps/vaults/hooks/usePortalsApi.ts
```diff
@@ -22,68 +25,62 @@ type TGetEstimateProps = {
...
 };
 
+const portalsTxSchema = z.object({
+	to: z.string(),
+	from: z.string(),
+	data: z.string(),
+	value: z.object({
+		type: z.string(),
+		hex: z.string()
+	}),
```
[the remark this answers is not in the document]
karelianpie: You're right, I just assumed all return values would be required, but they actually have the types there when you open the response 200

Screenshot 2023-06-12 at 11 50 34

Fixed that and others with https://github.com/yearn/yearn.fi/pull/230/commits/5ad309e137ebed3f785d9f5bac74c2770902498b

apps/vaults/hooks/useSolverPortals.ts
karelianpie: ^ @Majorfi

PR #232, opened 2023-06-12
fix: safe

Majorfi:
## Description
Fix an issue with safe: Safe require a valid bignumberish as value in order to work, while wagmi was sending undefined

## Related Issue
TG

## Motivation and Context
NA

## How Has This Been Tested?
Tested with a safe

## More
Misslabel the branch cowswap instead of safe


apps/common/utils/toWagmiProvider.tsx
```diff
@@ -72,7 +73,8 @@ export async function handleTx<
 		const config = await prepareWriteContract({
 			...wagmiProvider,
 			...props as TPrepareWriteContractConfig,
-			address: props.address
+			address: props.address,
+			value: toBigInt(props.value)
```
karelianpie, line 77: 👍

karelianpie (approved):

PR #246, opened 2023-06-20
fix: Add missing value to dependency array

Itzabelli:
## Description

The purpose to this pull request is to fix the warnings present in the repo, including the missing dependency.
After executing yarn lint you can observe the following warning in the terminal:

`⚠️ React Hook useCallback has a missing dependency: 'refetchAllowances'. Either include it or remove the dependency array.`

## Related Issue
N/A

## Motivation and Context

Adjust the dependency array to fix current warning and avoid code inconsistencies.

## How Has This Been Tested?

I confirmed that the warning was no longer present after making the change.

N/A


0xMirim (approved):
lgtm

0xMirim:
It looks like the repo lint rules have changed since the PR was made. These kind of missing dependency array values no longer trigger lint warnings. So it's probably safe to assume this will never be merged and we should close it 🥲

PR #278, opened 2023-07-13
fix: Use vaultsMigrations instead of vaults

karelianpie:
Use `vaultsMigrations` instead of vaults, as when this was implemented yDaemon was still old version


next.config.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/explicit-function-return-type */
 const runtimeCaching = require('next-pwa/cache');
-const withTM = require('next-transpile-modules')(['@yearn-finance/web-lib'], {resolveSymlinks: false});
+// const withTM = require('next-transpile-modules')(['@yearn-finance/web-lib'], {resolveSymlinks: false});
```
karelianpie, line 3: This `withTM` wasn't being used

PR #287, opened 2023-07-19
fix: Update .env.example file

Itzabelli:
## Description

I updated the .env.example file with the environmental variable: `WALLETCONNECT_PROJECT_ID`.

This change provides more clarity to the user and suggests setting this value to avoid errors on console.

## Related Issue

N/A

## Motivation and Context

Currently, this file doesn’t mention the need for the user to assign the value to this environmental variable. This may confuse the user.

## How Has This Been Tested?

Add the env file locally and assign the value of `WALLETCONNECT_PROJECT_ID`. 
After, I ran `yarn dev` and confirmed the error is no longer displayed.

N/A


karelianpie (approved):
LGTM 🚀

Issue #291, opened 2023-07-20
FRP: Let users search for both the vault name and symbol


0xBasically:
### What is your proposal 🚀
Let users search by both name and symbol of each vault
![image](https://github.com/yearn/yearn.fi/assets/131375535/e2690838-c3c4-4ed7-8173-fbd8e84121a8)

### What does this solve ? 🛠
![image](https://github.com/yearn/yearn.fi/assets/131375535/66100539-84b7-4c41-b5ed-da8fa5ec29ea)

PR #293, opened 2023-07-21
feat: Let users search for both the vault name and symbol

karelianpie:
## Description

Add the vault symbol to compare against the `searchValue`

## Related Issue

Closes https://github.com/yearn/yearn.fi/issues/291

## Motivation and Context

Let users search for both the vault name and symbol so they are more likely to find the vault they are looking for.

## How Has This Been Tested?

Tried locally, refer to the screenshot below

Screenshot 2023-07-21 at 17 07 03

### DIFF
--- a/pages/vaults/index.tsx
+++ b/pages/vaults/index.tsx
@@ -216,14 +216,13 @@ function Index(): ReactElement {
 	**	implemented as a simple string.includes() on the vault name.
 	**********************************************************************************************/
 	const searchedVaultsToDisplay = useMemo((): TYDaemonVault[] => {
-		const vaultsToUse = [...vaultsToDisplay];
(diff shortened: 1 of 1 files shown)


pages/vaults/index.tsx
```diff
@@ -216,14 +216,13 @@ function Index(): ReactElement {
 	**	implemented as a simple string.includes() on the vault name.
 	**********************************************************************************************/
 	const searchedVaultsToDisplay = useMemo((): TYDaemonVault[] => {
-		const vaultsToUse = [...vaultsToDisplay];
-
```
karelianpie, line 220: This seemed unnecessary, spreading the vaults to just put them back again into an array

PR #297, opened 2023-07-24
feat: boost display cleaning

Majorfi:
## Description
Change the display for the booster OPT vaults


karelianpie (approved):
LGTM! Potentially could be DRY'd a bit

PR #306, opened 2023-07-28
refactor: Include the strict typing syntax 'as const satisfies Chain'

Itzabelli:
## Description

Currently, the `wagmiNetworks.ts` file is made up of objects that we know always will have the specific typing `Chain` and that depending on the context its attributes can accept different values (number, string, etc..).

So, the purpose of this PR is to signalize the following syntax: `as const satisfies Chain` at the end of the each object, to ensure that the attributes and values that have at this time are kept and can't be changed.

This strict typing, can also be used to catch lots possible errors without needing additional verifications.

## Related Issue

N/A

## Motivation and Context

The intention is to apply the best practices in the case when we have a specific configuration that won't change.

Resources:
* https://wagmi.sh/core/chains#wagmicorechains
* https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator
* https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

## How Has This Been Tested?

N/A

N/A

### DIFF
--- a/apps/common/utils/wagmiNetworks.ts
+++ b/apps/common/utils/wagmiNetworks.ts
@@ -49,7 +49,7 @@ export const polygon = {
 		}
 	}
 
-};
+} as const satisfies Chain;
 
 export const optimism = {
 	...wOptimism,
@@ -88,4 +88,4 @@ export const optimism = {
 		}
 	}
 
-};
+} as const satisfies Chain;


karelianpie (approved):
LGTM! 🚀

PR #304, opened 2023-07-27
feat: improve wagmi usage

Majorfi:
## Description
Remove some old and deprecated behavior to focus on wagmi integration

PR #308, opened 2023-07-28
Feat/wagmi focus in weblib

Majorfi:
## Description
Follow up of #304 to move common stuff to the weblib

### DIFF
--- a/apps/common/utils/constants.ts
+++ b/apps/common/utils/constants.ts
@@ -1,13 +1,1 @@
-import {mainnet} from 'wagmi';
-import {arbitrum, fantom, optimism} from '@wagmi/chains';
-import {localhost} from '@yearn-finance/web-lib/utils/wagmi/networks';
-
 export const DEFAULT_SLIPPAGE = 0.5;
-
-export const SUPPORTED_CHAINS = [
-	mainnet,
-	optimism,
-	fantom,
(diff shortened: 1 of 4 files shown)


apps/common/utils/constants.ts
```diff
@@ -1 +1,13 @@
...
+
 export const DEFAULT_SLIPPAGE = 0.5;
+
+export const SUPPORTED_CHAINS = [
+	mainnet,
+	optimism,
+	fantom,
+	arbitrum,
+	localhost
+];
```
karelianpie, line 13: I would use a map here rather than an array, a bit more clear what we're getting `SUPPORTED_CHAINS.mainnet` rather than `SUPPORTED_CHAINS[0]`
```suggestion
export const SUPPORTED_CHAINS = {
	mainnet,
	optimism,
	fantom,
	arbitrum,
	localhost
};
```

pages/api/getBatchBalances.tsx
karelianpie, line 1: What's then reason for this file to exist? Can we just import getBatchBalances from '@yearn-finance/web-lib/utils/getBatchBalances' wherever it's used?

karelianpie (changes requested):
Looks good! A few comments

karelianpie (approved):

PR #316, opened 2023-08-07
chore: add op-weth and wusdrv2

dudesahn:
## Description

add info needed for yVOP rewards staking for recently added OP-WETH and wUSDRv2-USDC pools


karelianpie (approved):
LGTM

Issue #322, opened 2023-08-12
Bug: when enter a url of a vault with my wallet disconnected, the chain warning is wrong


saltyfacu:
### What happened 🆘
- I enter the URL of a vault.
- I see this: "Please note, this Vault is on Ethereum. You're currently connected to Ethereum."
- The vault I'm using is on Optimism

### To Reproduce 🔂
With your wallet disconnected, enter this URL: https://yearn.finance/vaults/10/0x531fDb1C8429F404116932226D487db94471C4b7

### More 🔎
- Browser :

PR #324, opened 2023-08-13
fix: Check if the wallet is connected and there's a current vault before showing toast

karelianpie:
## Description

Check if the wallet is connected and there's a current vault before showing toast

## Related Issue

Fixes https://github.com/yearn/yearn.fi/issues/322

## Motivation and Context

Show the warning correctly

## How Has This Been Tested?

Locally, screenshot below

Screenshot 2023-08-13 at 17 57 12


saltyfacu:
Nice! Thanks for the quick turnaround :)

Issue #323, opened 2023-08-13
Bug: zap on optimism not working


saltyfacu:
### What happened 🆘
I'm trying to zap into velodrome wusdr/usdc and the zap is not working. It says the Wido zap is not available, even when I select "portals". The warning appears 4 or 5 times.

### To Reproduce 🔂
- Go in https://yearn.finance/vaults/10/0x43360FDd9546e3e77ac2e6793f53219729743293
- Try to zap from USDC to the vault. Just writing the amount will do it

### More 🔎
- Browser : Chrome, Arc

karelianpie:
> Way to solve:
> 
> * Add a timer for that possible > 1 or 2s before a retrigger
> * Only trigger the toast after the user stopped typing for at least 1 or 2s
> * Support this zap

Ok, so I'll set the name of the current selected zapper in the toast and debounce it so we don't trigger it every time the user types in the amount input field

PR #325, opened 2023-08-15
refactor: Disable solvers in SettingsPopover dropdown

karelianpie:
## Description

* Disable solvers in SettingsPopover dropdown;
* Refactor the logic to display the toast;
* Refactor the `isSolverDisabled` to take the chain ID so we can have more complex logic regarding solver state;
* debounces the toast for wido, maybe worth to add this in the web-lib directly as an option perhaps.

## Related Issue

Fixes https://github.com/yearn/yearn.fi/issues/323

## Motivation and Context

We're allowing disabled solvers to be selected, and the toast is showing too frequently when changing the amount and the zap is not supported

## How Has This Been Tested?

Tested locally

Screenshot 2023-08-15 at 15 48 55


saltyfacu (approved):
Tested. It's working now. Thanks :)

PR #332, opened 2023-08-24
feat: Add useFeatureFlag hook

karelianpie:
## Description

* Implement `useFeatureFlag`

## Related Issue

Not directly related, but with this then we can merge https://github.com/yearn/yearn.fi/pull/248 and once ready remove the feature flag

## Motivation and Context

Allow for in progress features to be merged to reduce the need to keep open PRs up to date

## How Has This Been Tested?

Locally, added a value to the query param and was able to get a boolean out of it for the potential feature flag


karelianpie:
@Majorfi this is a pretty simple mechanism, let me know if you think we need something more complex

karelianpie:
> adding a "Corrupted Save Line ..." line

What's this again?

PR #336, opened 2023-09-01
chore: Add all missing peer dependencies

karelianpie:
A few packages require peer dependencies that were not met. We should however check that all these peer dependencies are even needed -> https://github.com/yearn/web-lib/blob/main/package.json#L102-L117

Also, upgrades web-lib to 2.0.0 🎉


karelianpie:
Closing in favour of https://github.com/yearn/yearn.fi/pull/337, added the missing deps there

PR #337, opened 2023-09-03
feat: update to lib-3

Majorfi:
## Description
Upgrade yearn-fi to webLib v3, currently on omega tag

## Related Issue
NA

## Motivation and Context
Cleaning and upgrading

### DIFF
--- a/apps/vaults/components/list/VaultsListRow.tsx
+++ b/apps/vaults/components/list/VaultsListRow.tsx
@@ -122,7 +122,7 @@ export function VaultAPR({currentVault}: {currentVault: TYDaemonVault}): ReactEl
 	);
 }
 
-function VaultsListRow({currentVault}: {currentVault: TYDaemonVault}): ReactElement {
+export function VaultsListRow({currentVault}: {currentVault: TYDaemonVault}): ReactElement {
 	const {safeChainID} = useChainID();
 	const balanceOfWant = useBalance(currentVault.token.address);
 	const balanceOfCoin = useBalance(ETH_TOKEN_ADDRESS);


apps/common/components/ListHero.tsx
```diff
@@ -170,4 +170,4 @@ function ListHero<T extends string>({
 		</div>
 	);
 }
-export default ListHero;
+export {ListHero};
```
karelianpie, line 173: I know it's a chore, but can we just export the function inline, rather than at the end of the file?

karelianpie (changes requested):
LGTM! Left a small comment/request

package.json
```diff
@@ -13,54 +13,55 @@
     "test": "vitest run"
   },
   "dependencies": {
-    "@cowprotocol/cow-sdk": "^2.1.0",
+    "@cowprotocol/cow-sdk": "2.1.0",
```
karelianpie, line 16: I was having some issues with this package with versions other than this one, so pinning it

package.json
```diff
@@ -69,27 +70,18 @@
...
-  "resolutions": {
-    "bn.js": "5.2.1",
-    "babel-loader": "9.1.0",
-    "framer-motion": "7.6.15",
-    "ethereumjs-util": "6.1.0",
-    "schema-utils": "4.0.0",
-    "tr46": "1.0.1",
-    "d3-color": "^3.1.0",
-    "json5": ">=2.2.2"
-  },
```
karelianpie, line 92: @Majorfi can you please double check that these are not needed? I removed all the resolutions, and AFAIK they don't seem to have any impact. Were these legacy ones or do you have some more context here?

karelianpie (approved):

Issue #340, opened 2023-09-10
Bug: bridged USDC is now called USDC.e


saltyfacu:
### What happened 🆘
The USDC vaults should be the USDC.e vault. The name changed, there's a native USDC version now.
We should change it everywhere.

### To Reproduce 🔂
Go here: https://yearn.fi/vaults/10/0xaD17A225074191d5c8a37B50FdA1AE278a2EE6A2
Check the token in the vault contract. The token name changed.

karelianpie:
Created X that renames USDC to USDC.e in the only place in the app that hardcoded the symbol name, all the rest comes downstream, @Majorfi is it hardcoded in yDaemon?

karelianpie:
Closing as there's nothing hardcoded in yearn.fi -- @Majorfi do you mind checking downstream?

karelianpie:
@saltyfacu will follow up here, the hardcoded label might be in yDaemon or even more upstream

0xBasically:
This is not a website issue but yDaemon issue

saltyfacu:
Created the issue in yDaemon: https://github.com/yearn/ydaemon/issues/331

PR #341, opened 2023-09-10
refactor: USDC -> USDC.e

karelianpie:
## Description

Change the name of USDC to USDC.e

## Related Issue

Closes https://github.com/yearn/yearn.fi/issues/340

## Motivation and Context

The USDC vaults should be the USDC.e vault. The name changed, there's a native USDC version now.


karelianpie:
> Are you sure about this one being here? The Native USDC is now on chain 10, here it's about chain 1

Ah, then we don't need this, didn't realise it was OPT only.

Issue #330, opened 2023-08-22
FRP: checkbox to not stake into vault


0xBasically:
### What is your proposal 🚀
Let users deposit into OP boosted vaults without staking if they choose. 

### What does this solve ? 🛠
Right now people can only deposit and stake into the staking contracts via our site, but if others would like to use vaults in parts of DeFi they need to not stake it but let others stake it for them. This will allow them to deposit into a vault and have the yvToken free to use elsewhere, if needed. 

### More 🔎
Make a checkbox on the site when staking to not stake it, with the default action being checked (staked).

karelianpie:
Waiting for @EvanKox's designs

karelianpie:
> Waiting for @EvanKox's designs

Decision here was to add this to the cog settings in the vault's detail page.

Now, looking at the contract, it appears we only have a zapIn function that deposits and stakes (https://optimistic.etherscan.io/address/0x498d9dcbb1708e135bdc76ef007f08cba4477be2#writeContract) -- we would need a similar function but without the staking part from the contract

https://optimistic.etherscan.io/address/0x498d9dcbb1708e135bdc76ef007f08cba4477be2#code#F1#L49

PR #345, opened 2023-09-12
refactor(Switch): Remove unknown

karelianpie:
## Description

* Add checkbox in the vault's detail page cog to not stake into vault (`SettingsPopover.tsx`);
* Add new local storage key/value pair `isStakingOpBoostedVaults`;
* Refactor logic to only use `OptimismBooster` if the user chose to stake automatically; otherwise continue with the action flow.

## Related Issue

Closes https://github.com/yearn/yearn.fi/issues/330

## Motivation and Context

Let users deposit into OP boosted vaults without staking if they choose. Right now people can only deposit and stake into the staking contracts via our site, but if others would like to use vaults in parts of DeFi they need to not stake it but let others stake it for them. This will allow them to deposit into a vault and have the yvToken free to use elsewhere, if needed.

## How Has This Been Tested?

Ran locally, refer to screenshots below.

Did not test with a real wallet.

### OP Boosted Vault

Checkbox is showing

#### Unchecked

Screenshot 2023-09-12 at 14 55 31

Button label is "Deposit"

Screenshot 2023-09-12 at 14 55 36

#### Checked

Screenshot 2023-09-12 at 14 55 41

Button label is "Deposit and Stake"

Screenshot 2023-09-12 at 14 55 45

### *NOT* OP Boosted Vault

Checkbox is not showing

Screenshot 2023-09-12 at 14 57 06


0xBasically:
Are you sure for op velo non-boosted vaults the checkbox doesn't show?
image

karelianpie:
> Are you sure for op velo non-boosted vaults the checkbox doesn't show?

Sorry, the description was a bit confusing. I wonder if that's what we want, or just show the checkbox for all OP vaults, people might try and look for the setting in a vault that is not boosted. So, for now I just added a check for the chain, do you think users will be able to find the setting if we only show it for boosted OP vaults?

0xBasically:
> > Are you sure for op velo non-boosted vaults the checkbox doesn't show?
> 
> Sorry, the description was a bit confusing. I wonder if that's what we want, or just show the checkbox for all OP vaults, people might try and look for the setting in a OP vault that is not boosted. So, for now I just added a check for the chain, do you think users will be able to find the setting if we only show it for boosted OP vaults?

I think we can grey it out making it not clickable if we dont have a staking contract for the vault

karelianpie:
> I think we can grey it out making it not clickable if we dont have a staking contract for the vault

After discussing with the team, we've decided to show the setting only for OP boosted vaults.

Done with https://github.com/yearn/yearn.fi/pull/345/commits/3bfd42c05dabffb430ce1530776d7458940fcc76

apps/common/components/Switch.tsx
```diff
@@ -1,36 +1,43 @@
...
 import {Switch as HeadlessSwitch} from '@headlessui/react';
+import {cl} from '@yearn-finance/web-lib/utils/cl';
 
-import type {ReactElement} from 'react';
+import type {KeyboardEvent, ReactElement} from 'react';
 
 type TSwitch = {
 	isEnabled: boolean;
-	onSwitch?: React.Dispatch<React.SetStateAction<boolean>>;
+	onSwitch?: () => void;
```
karelianpie, line 9: We don't really care what's the type of this function, simplifying

apps/common/components/Switch.tsx
```diff
@@ -1,36 +1,43 @@
...
-			onSwitch(!isEnabled);
-		} else {
-			set_isEnabledState(!isEnabledState);
+		onSwitch ? onSwitch() : set_isEnabledState(!isEnabledState);
+	}
+
+	const isChecked = onSwitch ? isEnabled : isEnabledState;
+
+	function handleOnKeyDown({key}: KeyboardEvent<HTMLButtonElement>): void | null {
+		if (key !== 'Enter') {
```
karelianpie, line 23: `keyCode` is deprecated in favour of `key`

apps/vaults/components/SettingsPopover.tsx
```diff
@@ -1,19 +1,31 @@
...
 import {useYearn} from '@common/contexts/useYearn';
 import {Solver} from '@common/schemas/yDaemonTokenListBalances';
 
 import type {ReactElement} from 'react';
 import type {TSolver} from '@common/schemas/yDaemonTokenListBalances';
+import type {TYDaemonVault} from '@common/schemas/yDaemonVaultsSchemas';
 
 type TSettingPopover = {
-	chainID: number
+	vault: TYDaemonVault
```
karelianpie, line 16: We need already 2 things from the current vault in this settings popover, so just passing the whole vault -- happy to refactor to just pass the `chainID`/`address` if strong opinions against this.

0xBasically (approved):
lgtm

karelianpie:
> LGTM code wise

Let's merge 🚀

PR #347, opened 2023-09-13
fix: Use the local copy of react

karelianpie:
In order to avoid having two copies of react when using the web-lib


karelianpie:
Closing in favour of https://github.com/yearn/web-lib/pull/334

PR #363, opened 2023-09-25
refactor: Remove outdated and unused redirects

0xMirim:
## Description

The redirects point to out outdated urls. Since there have been no complains I assume no one uses them and we can remove them. 

## Motivation and Context

Keep code tidy. Remove likely unused code providing limited utility.


karelianpie (approved):

0xMirim:
Just updated `/snapshot` redirect. Only remaining thing I wasn't sure about if is the first redirect is still valid or if we should remove it. Either way not a huge deal and we can probably just merge these changes as they are now.

karelianpie (approved):
LGTM!

PR #391, opened 2023-10-20
fix: factory part

### LINKED ISSUE #380: Bug: Cannot create new Curve factory vaults
### What happened 🆘
When user goes to create a new Factory Vault and click on "Available Curve pools", it doesn't load.

### To Reproduce 🔂
User 1:
Visit https://yearn.fi/vaults/factory, click on the "Select Curve Pool" arrow. If problem is persisting, nothing will load at this point.

User 2:
Same as user 1. See screenshot below:
image (1)

Console is full of errors referencing
`https://ydaemon.yearn.fi/1/vaults/all?hideAlways=true&orderBy=apy.net_apy&orderDirection=desc&strategiesDetails=withDetails&strategiesRisk=withRisk&strategiesCondition=inQueue
and 
https://ydaemon.yearn.fi/1/vaults/all?migratable=nodust

### More 🔎
User 1: Wallet is metamask, browser is Mozilla. 
User 2: Frame wallet, browser is Brave.

Majorfi:
## Description
Fix broken factory page.
Should fix #380

### DIFF
--- a/apps/common/contexts/useCurve.tsx
+++ b/apps/common/contexts/useCurve.tsx
@@ -1,4 +1,5 @@
 import {createContext, useContext, useMemo} from 'react';
+import {isZeroAddress} from '@yearn-finance/web-lib/utils/address';
 import {useFetch} from '@common/hooks/useFetch';
 import {coinGeckoPricesSchema} from '@common/schemas/coinGeckoSchemas';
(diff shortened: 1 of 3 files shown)


saltyfacu (approved):

Elvalandro (approved):
Fixed

PR #424, opened 2023-11-15
Feat/ext 5

Majorfi:
## Description
- Remove some extra hooks in the `MultiSelectDropdown`
- Refactor `useChain` to avoid rerender when `connectors` changes
- Fix `stop-opacity` to `stopOpacity`
- Allow `undefined` for `chains` or `categories` in `useVaultsQueryArgs`, where `undefined` is nothing selected. This will prevent all unselect to auto-reselect all

### DIFF
--- a/apps/vaults-v3/components/Filters.tsx
+++ b/apps/vaults-v3/components/Filters.tsx
@@ -9,11 +9,11 @@ import type {ReactElement} from 'react';
 import type {TMultiSelectOptionProps} from '@common/components/MultiSelectDropdown';
 
 type TListHero = {
-	categories: string[] | undefined;
-	chains: number[] | undefined;
+	categories: string[] | null;
+	chains: number[] | null;
 	searchValue: string;
-	onChangeCategories: (categories: string[] | undefined) => void;
-	onChangeChains: (chains: number[] | undefined) => void;
+	onChangeCategories: (categories: string[] | null) => void;
+	onChangeChains: (chains: number[] | null) => void;
 	onSearch: (searchValue: string) => void;
 };
 
--- a/apps/vaults/components/ListHero.tsx
+++ b/apps/vaults/components/ListHero.tsx
@@ -9,12 +9,12 @@ import type {TDict} from '@yearn-finance/web-lib/types';
 import type {TMultiSelectOptionProps} from '@common/components/MultiSelectDropdown';
 
 type TListHero = {
-	categories: string[] | undefined;
+	categories: string[] | null;
 	possibleCategories: TDict<string>;
-	chains: number[] | undefined;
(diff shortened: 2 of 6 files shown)


apps/vaults-v3/components/Filters.tsx
```diff
@@ -9,11 +9,11 @@ import type {ReactElement} from 'react';
 import type {TMultiSelectOptionProps} from '@common/components/MultiSelectDropdown';
 
 type TListHero = {
-	categories: string[];
-	chains: number[];
+	categories: string[] | undefined;
+	chains: number[] | undefined;
```
karelianpie, line 13: How come explicit `undefined`? Should we do one of these instead:
```suggestion
	categories?: string[];
	chains?: number[];
```
or
```suggestion
	categories: string[] | null;
	chains: number[] | null;
```

apps/vaults-v3/components/Filters.tsx
```diff
@@ -9,11 +9,11 @@ import type {ReactElement} from 'react';
...
 	searchValue: string;
-	onChangeCategories: (categories: string[]) => void;
-	onChangeChains: (chains: number[]) => void;
+	onChangeCategories: (categories: string[] | undefined) => void;
+	onChangeChains: (chains: number[] | undefined) => void;
```
karelianpie, line 16: Same as https://github.com/yearn/yearn.fi/pull/424/files#r1394103146

apps/vaults-v3/components/Filters.tsx
karelianpie: or even better, not have these as neither `undefined` nor `null` but rather an empty array when there're no categories nor chains. Wdyt?

apps/vaults/components/ListHero.tsx
```diff
@@ -53,7 +53,7 @@ export function ListHero({
 							const selectedChains = options
 								.filter((o): boolean => o.isSelected)
 								.map((option): number => Number(option.value));
-							onChangeChains(selectedChains);
+							onChangeChains(selectedChains.length === 0 ? undefined : selectedChains);
```
karelianpie, line 56: I'm not sure this is a good change, adds complexity for no apparent reason.

I would try and have the empty array, or why is that not possible and we need to go with `undefined`?

apps/vaults/hooks/useFilteredVaults.ts
karelianpie, line 1: The changes in this file 
image

karelianpie (changes requested):
I would try and not have the `undefined` stuff, rather have the empty array

karelianpie (approved):
Let's go with this approach 🚀

PR #639, opened 2025-03-25
feat: deposit only according to solver and live rewards

Karlen9:
## Related Issue

(https://github.com/users/rossgalloway/projects/3/views/2?pane=issue&itemId=99199273&issue=rossgalloway%7CySHIP%7C46)


w84april (approved):

rossgalloway:
Something wrong here. this APY is higher than the max. This is without a wallet connected.

image

With a wallet connected to an account with veYFI I get this:

image

other vaults with rewards still show ranges (with wallet connected):

image

rossgalloway:
Still a few issues:

- When switching the toggle from deposit and stake to just deposit, the APY for only deposit includes the 1x boost amount (it shouldn't).
- text under "from wallet box" not aligned with text under other boxes. may only be an issue on firefox (?)
image

PR #640, opened 2025-03-25
Feat/unstake amount

w84april:
image
and unstake amount is editable:
image


rossgalloway:
When trying to unstake:

image

image

both for max and an entered value.

w84april:
@rossgalloway fixed [here](https://github.com/yearn/yearn.fi/pull/640/commits/ef07e0cd6880b3ee80233eef333453a1e7a9e6c5)

PR #644, opened 2025-04-09
feat: init new strats

w84april:


rossgalloway:
I think this is mostly good to go. It updates the strategies tab to have more information and looks like what I have on powerglove

murderteeth (approved):

PR #645, opened 2025-04-21
feat: display correct symbol

w84april:
## Description

Display gauge asset symbol if "Deposit & Stake" is on

## Related Issue

https://github.com/rossgalloway/ySHIP/issues/46


rossgalloway:
I think this is good to go. pretty simple.

murderteeth (dismissed):
lg - note on multicalls. we should start using key-value maps instead of bare result arrays. the difference is in not having to document and use indexes on the results, we use readable named keys instead. maintainability+

rossgalloway:
> lg - note on multicalls. we should start using key-value maps instead of bare result arrays. the difference is in not having to document and use indexes on the results, we use readable named keys instead. maintainability+

new commit attempts to fix this. How did 04-mini do?

murderteeth (dismissed):

PR #648, opened 2025-05-02
Chore: update /apps page

rossgalloway:
## Description

* remove gimme from featured
* remove juiced
* add factory to vaults
* remove yprisma from yearn apps
* put juiced, yPrisma in "old apps"
* remove search bar
* add yprisma gauge to disabled list

## Related Issue

To make FP happy

## Motivation and Context

FP

## How Has This Been Tested?

locally running. no unit tests.


rossgalloway:
featured products in mobile only shows one item, and it is not clear that there are others. Change featured products to be stacked banners instead

rossgalloway:
- add old products to page when in mobile since it is hard to find otherwise (buried in hamburger menu)
- add swipe-able second group of links when there are more than 4 items in a category. Add dots at bottom to show position. Currently only applies to Yearn X Projects

w84april:
Looks nice in both themes
Just 2 things I'd like to mention:

1. I really like old border color and hover color, wdyt of using it instead of new one
image
image

2. We probably don't have to render border on mobile:
image

murderteeth (changes requested):
https://github.com/user-attachments/assets/ac766729-31a5-4585-a5aa-d693ceb7e978

i notice a lot of "layout shift" on the route transition animation, looks janky to me. fwiw i dont like large\dreamy animations for route transitions like this. i'd be just as happy to get rid of them entirely fn vs fix the shift. the flow would feel faster and cleaner i think

rossgalloway:
Removed all motion from page transitions. Kept the subtle fade in and out. Back button on v3 vault pages changed from router.back to instead just navigate to the /v3 page, which stops the weird scroll effect.

PR #649, opened 2025-05-08
Chore: add sonic

rossgalloway:
## Description

Add required code to display sonic vaults.

## Related Issue

https://github.com/rossgalloway/ySHIP/issues/72

## Motivation and Context

Need to display sonic vaults on the front-end

## How Has This Been Tested?

none yet

### DIFF
--- a/apps/common/utils/constants.tsx
+++ b/apps/common/utils/constants.tsx
@@ -1,3 +1,4 @@
+import {defineChain} from 'viem';
 import {arbitrum, base, fantom, mainnet, optimism, polygon} from 'viem/chains';
 import {toAddress} from '@builtbymom/web3/utils';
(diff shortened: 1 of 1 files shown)


rossgalloway:
Problems: I see Sonic as an option on the V2 page but not on V3. I don't really know where the issue is. There may be an issue with wagmi as well? I had to add sonic manually since the version of viem that the builtbymom package is using it doesn't include it (too old). This may also be true for wagmi. Only hint I have is that when I try to navigate the sonic vault that Johnny deployed `yearn.fi/v3/146/0xb9228370e2fa4908fc2bf559a50bb77ba66fdd66` I get an error that chain 146 is not supported that seems to stem from a wagmi function.

PR #650, opened 2025-05-13
chore: update curve URLs

rossgalloway:
## Description

update curve.fi to curve.finance after they migrated due to a DNS hack

## Related Issue

none

## Motivation and Context

Keep site working correctly

## How Has This Been Tested?

not tested


rossgalloway:
This should be merged after yDaemon is updated to change its URLs to curve.finance

murderteeth (approved):

PR #657, opened 2025-05-19
fix: strat filtering

w84april:
## Description

Multiple improvements related to new strategies view mostly, here's the list:
- Use `recharts` lib in favor of custom solution for PieCharts
- PieChart has now different styling (pinkish for v3 and greyish for v2)
- Display unallocated amount in chart
- Display unallocated amount in list + calculate unallocated amount in $
- Add `Show/Hide unallocated` button and unallocated strategies view
- Make `Vault` param unsortable for now as it doesn't do anything for now and pretty useless for this view imo (need research, smth might be broken)
- Fix APR display for most of the v2 strategies (it was >500 most of the times)
- Remove duplicated VaultsV3ListStrategy.tsx
- Make sorting work for unallocated vaults as well
- Remove extra strategy filtering
- Fix typos

It is mentioned in https://github.com/rossgalloway/ySHIP/issues/83?issue=rossgalloway%7CySHIP%7C82: `add the color to the label that shows when you hover`. Skipped this for now, not sure how to address and feels a bit extra as it is very clear what pie is being hovered. Lmk if i miss smth

## Related Issue
https://github.com/rossgalloway/ySHIP/issues/83

## Screenshots
image

The way `unallocated` amount is shown in the list:
image

 
v2:
image
image


rossgalloway:
nice! Good progress here.

A few things on first glance:
- Some of the strategy names are missing. These are new strategies, so maybe it is a yDaemon issue? But if you could take a look. 
- I like the pink on the chart, but the pink "show unallocated" button should be much more subtle. It should feel like a part of the list and not attract attention. Also, lets change the name to "show inactive strategies". That will differentiate it from "unallocated funds".
- Unallocated funds should always be the last entry on the list of active strategies.
- The chart should always go above the unallocated vaults. So when the chart drops to the bottom, maybe put the "show inactive strategies" button underneath it.
- The spacing of the strategies data is broken in mobile. maybe we stack the data when mobile sized.
    image

- The links when you expand to see more data overflows the component when in mobile. Let's get rid of the chain element in all views and that should fix it.
    image

- the color for the chart label is un-needed as you mentioned. I assumed all the strategies would show on the popover, but they don't. What you have is good.

w84april:
@rossgalloway Thanks! Pushed more stuff related to the things you described, should be better now. 

> Some of the strategy names are missing. These are new strategies, so maybe it is a yDaemon issue? But if you could take a look.

Yup, that's meta in yDaemon. Will take a look tomorrow

Some screenshots:
Hide/show button is much more subtle:
image

Chart is between two lists:
image

Mobile view is adjusted:
image

V2 variant:
image

rossgalloway:
LGTM. @murderteeth needs an approving review, which it doesn't seem like I can give.

PR #663, opened 2025-05-26
chore: fork lib

w84april:
## Description

Not possible to publish prev version of web-lib anymore, so forked it. Tried linking git repo directly but it was pain.
- Now using https://github.com/w84april/web-lib instead of https://github.com/yearn/web-lib (didn't update imports, just added tsconfig rule)
- Bumped @builtbymom/web3, wagmi, viem
- Bumped node in git workflow

## Related Issue
fixed here - https://github.com/rossgalloway/ySHIP/issues/85
fixed here - https://github.com/rossgalloway/ySHIP/issues/86
fixed here - https://github.com/rossgalloway/ySHIP/issues/87
tldr re 87: not all strategies were vaults (had to hide some) and vault page -> vault page redirect didn't work in the same tab. Quick fix is just to make it open in new tab which works perfectly imo


rossgalloway:
I'm getting errors in the preview build when trying to view vault pages. We have similar errors in https://github.com/yearn/yearn.fi/pull/652, which also has a forked mom library, so they may be related.

image

image

PR #661, opened 2025-05-23
feat: new landing page

0xeye:
## New homepage. Gm.

This is part one of a wider UX re-focus and product consolidation, see [here](https://github.com/yearn/budget/issues/290) for initial scope.

This PR aims to:
- Improve Yearn landing page and discovery
- Create simpler user journeys (re getting users to key products in fewer steps)
- Bring in data to improve liveness of the site

Next steps will be to radically cut the https://yearn.fi/apps page, making the focus solely Vaults (v2 & v3). This will come in a later PR. 

## Motivation and Context

Yearn has great products, and great numbers. Current landing page, in my opinion, does not shout about these loud enough.

## How Has This Been Tested?

Yes, but likely there will be some adjustments needed. 

|| || **Please feedback on copy** || || 

## Screenshots (if appropriate):

| Current  | New |
| ------------- | ------------- |
|  Screen Shot 2025-05-23 at 16 27 10-fullpage |  Screen Shot 2025-05-23 at 17 16 41-fullpage |

PR #672, opened 2025-06-02
feat: consolidated vaults

0xeye:
## Consolidated Vaults.

This is part one of a wider UX re-focus and product consolidation, see [here](https://github.com/yearn/budget/issues/290) for initial scope.

This PR aims to:
- Consolidate V2&V3 vaults and improve UX

The base of this is the landing page PR (#661 ) so should be merged subsequent to that being merged

## Motivation and Context

V2 and V3 are both active vaults and users shouldn't be sent to different pages.

## How Has This Been Tested?

Yes, but likely there will be some adjustments needed. Preview PR for now

| Main  |
| ------------- | 
| Screen Shot 2025-06-02 at 16 08 17-fullpage |


murderteeth:
* mobile landing page - menu icon is still a text-justify icon
- mobile - prefere menu button at top left, thats where mobile users most expect it
- mobile vaults - header feels cluttered with wallet right next to menu. users like clear and easy tap targets
- mobile - if we're committed to top-left branding, lets try the Y instead of a hamburger
- mobile - still think its better to move landing page link to menu instead of in the header
  - top left of page = most important
  - but how often do users want to go back to the landing page?
  - and is that what we want users doing after entering? instead of navigating literally anywhere else in the site
* mobile vaults - looks like filters need more layout work. they're pinched\overflow for me
- mobile vaults - chain selector (and similar components). when its opened, instead of a dropdown, let it take full screen. much easier to use this way on mobile

PR #673, opened 2025-06-03
Fix/strategies

w84april:
## Description

Some strat fixes

## Related Issue

https://github.com/rossgalloway/ySHIP/issues/85
https://github.com/rossgalloway/ySHIP/issues/86
https://github.com/rossgalloway/ySHIP/issues/87

## Screenshots
image


rossgalloway:
Re: https://github.com/rossgalloway/ySHIP/issues/87

Looks like you changed the functionality to open the vault as a new tab. The request I got was to remove the icon showing that it opens in a new tab, and keep it so that it opens the page in the current tab. Let's do that since it stays within our site.

PR #678, opened 2025-06-05
Chore: add sonic (for real this time)

rossgalloway:
Replaces the other PRs for sonic (#651 and #652) and implements their changes (plus a few more) on top of the newest code.

- We should consider a better way to add chains to the site as there are currently 3-4 places where they need to be manually added for everything to work right.
- Another little quirk that required a work-around is that the sonic logo from smol is black on transparent so basically invisible in v3 and v2 (with dark move). I added a white background to all chain logos in the dropdown to fix this. Side effect is there is a tiiiny white outline around the logos now.

This PR also incorporates some other small adjustments from prior PRs that can now be closed. ( #650 and #659 ).

### DIFF
--- a/apps/lib/components/MultiSelectDropdown.tsx
+++ b/apps/lib/components/MultiSelectDropdown.tsx
@@ -57,7 +57,7 @@ function Option(option: TMultiSelectOptionProps): ReactElement {
 				onMouseLeave={() => set_isHovered(false)}>
 				<div className={'flex items-center'}>
 					{option?.icon ? (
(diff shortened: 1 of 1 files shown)


murderteeth:
> * Another little quirk that required a work-around is that the sonic logo from smol is black on transparent so basically invisible in v3 and v2 (with dark move). I added a white background to all chain logos in the dropdown to fix this. Side effect is there is a tiiiny white outline around the logos now.

instead of a bg, try https://tailwindcss.com/docs/filter-invert and or https://tailwindcss.com/docs/filter-brightness

rossgalloway:
I broke the linter check with my branch name... won't use parentheses again

PR #685, opened 2025-06-10
fix: v2 balances

w84april:
## Description

There was a bug in balances filtering: if vault has staking balance, but not balance, it wasn't taken into consideration in sum amount.
Also removed some unused code to keep things cleaner

## Related Issue

https://github.com/users/rossgalloway/projects/3/views/2?pane=issue&itemId=111159556&issue=rossgalloway%7CySHIP%7C78


w84april:
@rossgalloway That were a bit tricky ones! Pushed fix:
1. Instead of including Pendle vaults into total deposited, we had to hide them from v2 vault list as they are v3, so had to update filtering logic a bit:
image
2. `cumulatedValueInVXVaults` didn't actually take into consideration any vault other than active
3. Same stuff with `balances` calculation, now migrations and retired vaults are included

rossgalloway (approved):

PR #690, opened 2025-06-23
Chore/optimize balances

w84april:
## Description

- Move balances logic out of useYearn
- Implemented React's useTransition hook to prevent UI blocking during balance fetches
    - Accumulates all balance updates and applies them in a single batch
    - Returns empty balances during pending state to prevent UI from freezing (idk that's probably my laptop issue but m1 max should be good enough to handle react app)

### DIFF
--- a/apps/vaults-v3/components/details/actions/QuickActionsFrom.tsx
+++ b/apps/vaults-v3/components/details/actions/QuickActionsFrom.tsx
@@ -1,7 +1,5 @@
 import {useCallback, useMemo} from 'react';
 import {useRouter} from 'next/router';
-import {erc20Abi} from 'viem';
-import {useReadContracts} from 'wagmi';
 import {useActionFlow} from '@vaults-v2/contexts/useActionFlow';
 import {Renderable} from '@lib/components/Renderable';
 import {Dropdown} from '@lib/components/TokenDropdown';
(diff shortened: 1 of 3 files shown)


murderteeth (changes requested):
- clean up console.logs
- is there a worker? the comments mention one but i dont see it. if there's no worker, please clear up the documentation too
- are you sure useTransition makes a difference here? useTransition is designed for cases where the ui update itself is large. for example updating a table with a lot of rows.
- do you feel we could refactor this into less code? maybe using tanstack for example. my intuition is that we could. but what do you think? doesn't have to be this pr

murderteeth (changes requested):
image

should my balance equal the sum of my holdings? maybe a bug

w84april:
thanks @murderteeth 
- cleaned up all worker mentions for now (might actually add one to test later)
- yep, useTransition is removed
- don't think tanstack query might be useful here, but I did refactor main `useAsyncTrigger`, `onUpdate`, `onUpdateSome` functions that were doing almost same stuff. Now they all just call the same func with slightly different params. Also actually made `onUpdate`, `onUpdateSome` work as expected (force refetch balances and update balances state). Removed one more `useContractCall` that was triggered after withdraw/deposit and was doing their job  
- as for https://github.com/yearn/yearn.fi/pull/690#pullrequestreview-2956080214 - the issue is that [this prisma vault](https://etherscan.io/address/0x11AaE8beE9b1Da827C641540D20e4e664677e06F) is calculated in total assets but not displayed in list bc we don't have `Prisma` category. So not sure what to do here, mb safe to ignore? We are calculating total deposited value out of active, migratable and retired vaults, so changing it's status won't help

image

murderteeth (approved):
lgtm

> the issue is that [this prisma vault](https://etherscan.io/address/0x11AaE8beE9b1Da827C641540D20e4e664677e06F) is calculated in total assets but not displayed in list bc we don't have Prisma category. So not sure what to do here, mb safe to ignore?

confirmed, i tested w a different wallet and balances matched. yeah maybe we need a retired category that only shows if you have a retired balance. ask in the group. but agree, this issue isn't related to the pr

PR #704, opened 2025-07-17
chore: misc fixes

0xeye:
Fixes in order of commits:

- https://github.com/rossgalloway/ySHIP/issues/114
- https://github.com/rossgalloway/ySHIP/issues/112 
- https://github.com/rossgalloway/ySHIP/issues/111
- https://github.com/rossgalloway/ySHIP/issues/116

### DIFF
--- a/apps/lib/components/RenderAmount.tsx
+++ b/apps/lib/components/RenderAmount.tsx
@@ -27,6 +27,7 @@ export function RenderAmount(props: TAmount & {shouldHideTooltip?: boolean; shou
 		((props.value < 0.001 && props.symbol !== 'percent') || (props.value < 0.0001 && props.symbol === 'percent'));
 
 	const value = formatTAmount(props);
+
 	return (
 		<span
 			title={value}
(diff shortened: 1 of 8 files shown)


rossgalloway:
The Dust Guard implemented to fix https://github.com/rossgalloway/ySHIP/issues/111 is a big improvement but highlighted another issue.

If there is dust in a vault, it populates to the top of the list, even if not considered active:

image

Note in the above, the DAI vault is listed above the USDS vault because of the holdings. The default looks like this:

image

I think the solution here is to more agressively cull dust based on USD amounts. If less than $0.01 in USD, we should round to 0. 

(Side note: We may want to start thinking about showing values in USD as the default instead of how we currently show them in the underlying asset. This would allow us to be consistend in formatting decimals and will fix things like "4 decimals is too much for stables, but too few for WBTC".)

0xeye:
I've adjusted the filter from $0.001 to $1 value to prevent lower values from showing in the top section. I think the dust problem should feature less with the fix in https://github.com/yearn/yearn.fi/pull/655 but lets spin that out into another ticket & get these in for now imo.

PR #732, opened 2025-08-27
feat: enable search for strats

w84april:
## Description

Make strategy name, strategy address, underlying token address searchable 

## Related Issue

No issue, chat request


w84april:
> I don't think this solves the issue. From what I can tell, this searches vaults for new fields: strategy names and addresses. This means it now returns vaults that contain these strategies.
> 
> But the issue is that these strategies are tokenized strategies (still vaults, like the morpho strategies) and should show up as vaults when searched (when the filters are properly selected). We see in this PR that even though the vaults that contain these strategies show up, the strategies themselves do not.
> 
> Another quality of life change that we should try to improve as well is that when searching for something, and it is blocked by a filter, we should still show the search results, but alert the user that it is not part of the current filter group.

Makes sense, checking

PR #733, opened 2025-09-01
fix: display full list

w84april:
## Description

Only fancy vaults are currently displayed by default. Probably makes sense to display all of them

before:
image
after: 
image


rossgalloway:
Based on discussions I have had with strategists, They would like to default to fancy and not show strategies by default.  

As discussed, we should make it easy to search for strategies and vaults that aren't in "fancy", but not have them in that initial list.

PR #737, opened 2025-09-04
Feat  add steer rewards

rossgalloway:
## Description

- This PR builds on the updates to the katana-apr-service API upgrade to show Steer points. 
- Adds a contextual tooltip showing “Eligible for Steer Points” for Katana vaults in the list view when there is a campaign, Explains registration and points earning in the tooltip and in the existing APR modal. The tooltip is styled consistently with existing tooltips and renders below the label.
- Updates the styling of the Katana APY tooltip to be easier to read and better match what is on katana.yearn,space
- greys out the bonus APR as that is only valid if a user deposited in the vault before sept 1st. It is also not included in the calculation per request from katana team.
- This modal also exists on the vault page when hovering over "APR Breakdown"
- Also adds an AGENTS.MD file to the repo.

Ideally, this PR is approved after https://github.com/yearn/katana-apr-service/pull/11 since that is where the new data comes from. Without it, the site will still work.

## Related Issue

https://github.com/yearn/katana-apr-service/pull/11
https://github.com/yearn/yearnX/pull/44

## Motivation and Context

Show Steer Points and Clean up the APY breakdown modal.

## How Has This Been Tested?

locally tested

image


murderteeth (approved):
I think its ok to fix this in another PR, but noting that our APR hovers are starting to look inconsistent.

Image

Image

PR #740, opened 2025-09-09
Upgraded App [Vite]

0xeye:
Changed project from NextJS -> Vite & made relevant changes to support this shift.


rossgalloway:
> mobile menu is off

I can pick this up with my mobile tooltips changes

murderteeth (changes requested):
plausible integration is broken

Image

murderteeth (changes requested):
landing page load performance has also changed. this was more obvious on a Windows machine with a vertical scrollbar. we can maybe merge and fix load performance after. but please take a took
490569857-e195b192-606f-4a81-9ecc-0eaa775f3649-ezgif com-optimize

PR #745, opened 2025-09-15
Route pre-fetching

rossgalloway:
## Description

I wasn't happy with the responsiveness of the route changes, so I pointed codex at it, and it spat out some pre-fetching. Seems to make a difference
Also added an AGENTS.md file

## Motivation and Context

Must be snappy with no transitions

## How Has This Been Tested?

locally


0xeye:
Don't think improves speed much. Perhaps better solution is to just have a progress spinner when loading content (would also improve flow for cold landing to vaults pages)

PR #746, opened 2025-09-16
Fix/list v3

w84april:


w84april:
> What are the "strange holdings behaviors" in the screenshot?

Felt wrong to have dusty vaults appear in non-holdings, but now thinking different approach should be here. We should still keep them out of holdings, but display zeros instead of what we have on screenshot, feels buggy
image

PR #752, opened 2025-09-21
test: force refetch

w84april:


murderteeth (changes requested):
- balances under input token are updating, but balances under staking are not. first two images below show what ybold looks like immediately after a deposit and immediately after a withdraw. after deposit i expect available to stake to be equal to shares. after withdraw i expect available to stake to be zero.

- it looks like these changes introduced a react performance bug. see video. on left side is the preview running off pr740, on the right is the preview running off pr752. in both cases i'm just going back and forth on the deposit withdraw tabs. if you pay attention specifically to the label on the deposit\withdraw button, you'll see what I mean

- remove bun.lock (easier to merge)


Image

Image

https://github.com/user-attachments/assets/f00d37a7-e7db-42cc-b8ff-31564e4459e4

0xeye (approved):
LGTM! Do we need any awaits on the async calls? Otherwise good to me

PR #747, opened 2025-09-16
Fix  mobile modals

rossgalloway:
## Description

- Fix modals on mobile. In the V3 landing page, they now expand the card. APY and risk score tooltips implemented. 
- Update mobile VaultV3ListRow card to remove holdings (unless you have holdings) and available. restyled a little bit to clean up the card. network up top, show risk score better.
- updated the token image fetching to all use the CDN and read from the .env file for the url.
- Fixed the mobile hamburger menu and moved the hamburger to not trigger the modal. (should probably deactivate the nav modal in a future PR)

## Related Issue

https://github.com/yearn/yearn.fi/pull/740
original tooltip upgrade here: https://github.com/yearn/yearn.fi/pull/737

## Motivation and Context

Make it better

## How Has This Been Tested?

locally

## Screenshots (if appropriate):

Screenshot 2025-09-16 133240
Screenshot 2025-09-16 133325
Screenshot 2025-09-16 133302
image

PR #757, opened 2025-09-25
Fix: add additional katana info to tooltip

rossgalloway:
## Description

- Updated the katana tooltips to include information on FDV, unlock date, and a link to more info. This now matches info on the katana.yearn.space site.
- cleaned up the tooltips to make sure they render correctly on mobile.

## Related Issue

#747 

## Motivation and Context

Make things better. Requested in Discord

## How Has This Been Tested?

locally

image
image


murderteeth (dismissed):
- small thing, there's a click trap to the left of the apr. if you tap there nothing happens. tap above or below the trap or right on the apr itself and works as expected.  
- on mobile i dont see a way to open the apr tooltip on the actual vault view (if there's supposed to be one)

Image

PR #767, opened 2025-10-01
fix: deposit and stake

w84april:
## Description

Few bugs:
1. Wrong decimals in Unstake section
2. Wrong `selectedOptionTo` token when `Deposit and Stake` is on (GaugeStakingBooster). Previously it was just vault token with correct symbol rendered, now it's staking token.


apps/vaults-v2/hooks/solvers/useSolverGaugeStakingBooster.ts
```diff
@@ -1,19 +1,43 @@
...
+  chainID: number
+  outputToken: TAddress
+  inputToken: TAddress
+  inputDecimals: number
+  outputDecimals: number
+  isDepositing: boolean
+  version: string
+  from: TAddress
+}
+const getEstimageOut = async (request: TGetVaultEstimateOutProps): Promise<TNormalizedBN | undefined> => {
```
rossgalloway, line 31: this should be named `getEstimateOut` and not `getEstimageOut` right?

PR #773, opened 2025-10-07
feat: nextgen hooks

0xeye:
Adds new helper functions & components to gradually move logic toward wagmi hooks + TxButton. 

Would be good to get this in & then we can rebase the widgets branch and continue work there re solvers/zaps


rossgalloway (approved):
LGTM

PR #776, opened 2025-10-08
Feat/widget solvers [WIP]

w84april:
## Description

Some more logic for new widgets


rossgalloway:
you may already know this, but the widget it not auto-switching the wallet like the rest of the app.

image

Requires going into the wallet and changing the chain

rossgalloway:
also, the dropdown pulls from available assets but then when the default asset is changed it is no longer available.

https://github.com/user-attachments/assets/4009644b-3a8e-4bf8-afce-b415658563f2

w84april:
changes are merged into https://github.com/yearn/yearn.fi/pull/838

Issue #789, opened 2025-10-13
Feature Request: Header logo opens modal navigation for Yearn properties


rossgalloway:
### Summary
- Replace the current hover-only popover triggered by the Yearn header logo with a click-to-open modal that lists Yearn apps and internal tools as tiles.
- Preserve the existing app shortcuts and add tiles for the landing page, `oracle.yearn.fi`, `cms.yearn.fi`, `token-assets.yearn.fi`, `https://yearn.space` and `kong.yearn.fi`; ensure integrations are excluded.

### Background & Motivation
- The existing popover (`apps/lib/components/LogoPopover.tsx`) is hard to discover, requires hover on desktop, and is awkward on touch devices.
- Stakeholders want the header logo to become a richer discovery surface for both public apps and useful internal tooling, without adding extra hops between landing pages and primary app flows.
- The "Apps" page is deprecated, but it is still useful to have a place where all the different useful yearn apps live.

### Acceptance Criteria
- Clicking the Yearn logo on any page opens a modal (using an accessible dialog with focus trap and overlay) instead of navigating home immediately.
- The modal displays all existing app entries from `apps/lib/components/YearnApps.tsx` (V3 Vaults, Vaults, yCRV, veYFI, yFactory, Analytics, yETH, yPrisma, GIMME, Vaults Beta) and adds tiles for:
  - Home — `https://yearn.fi/`
  - Oracle — `https://oracle.yearn.fi`
  - CMS — `https://cms.yearn.fi`
  - Token Assets — `https://token-assets.yearn.fi`
  - Seafood — `https://seafood.yearn.watch`
  - Kong — `https://kong.yearn.fi`
  - Also add tiles for other items in the header: docs, support, blog, discourse.
- Each tile shows an icon (fallback to Yearn glyph if no bespoke asset), title, short description/label if desired, and opens the destination in a new tab if it leaves `yearn.fi`.
- Provide a clear “Home” tile or button linking to `/` since the logo no longer navigates directly home.
- Modal supports keyboard/assistive tech: focus moves inside the dialog, `Esc` closes it, and there is an explicit close button or click-away handler.
- The hover-state animation that indicates the active app in the header remains functional (or is replaced with an equivalent indicator inside the modal).
- Remove the “More…” accordion logic; all links are visible without extra clicks.
- Remove or refactor unused hover state management, framer-motion animation code, and `Popover` dependencies no longer needed.
[...]

PR #793, opened 2025-10-15
Feat/logo modal clean

rossgalloway:
## Description

Create new nav modal to replace existing and apps page.

This PR supercedes #790 and incorporates the changes made to that, but points at the main branch to keep this work separate from other UX and design changes.

- [x] set "apps" link on landing page to open modal
- [x] fixed light and dark mode styling
- [x] freeze\disable page scroll when modal is open
- [x] re-organized vault order
- [x] Bearn and Kalani added
- [x] Changed component name to "launch modal".
- [x] Re-factored modal so there is a side menu with categories and all options fit nicely. Apps are focused. 
- [ ] Improved Icons, but not all. Due to the refactor this isn't as blatantly bad since they aren't all seen together. Can improve but probably not a blocker. 

## Related Issue

https://github.com/yearn/yearn.fi/issues/789

## Motivation and Context

make shit better

## How Has This Been Tested?

locally

image
image
image


murderteeth (changes requested):
-- unify button style
group buttons and launch buttons should share the same interaction language. it feels disjointed otherwise. my preference is for the styling on the launch buttons. regardless, same is better than different.

specfically, 
- use same border radius
- use same transition effect, (ie none)
- use same general interaction pattern: change style on hover, mousedown, and mouseup\out

apply same pattern to the Y launcher icon. this makes the path from hovering the Y to launching an app feel continuous

-- use the whole screen on mobile
use fullscreen for mobile dialogs — floating modals create a cramped ux. another option is the bottom-sheet pattern (eg rainbow wallet). but for a dialog with many options, fullscreen is better

-- on mobile, h-scroll is broken for the group buttons surface. it shows an h scrollbar (👎) and doesn't seem to support drag. i have used https://www.npmjs.com/package/react-indiana-drag-scroll in the past. but its an old dep with not much code. ai could probably recode it for us instead. another option: no group button hscroll on mobile. revert to previous layout, all groups in once view, simple fullscreen vscroll.

-- nicetohave: a dark\lite toggle. instead of figuring out a button now, use an easteregg: maybe a hotkey or just double-clicking the Y
-- nicetohave: hamburger and notifications should share the same modal theme w launcher. otherwise we now have 4 competing modal themes: launcher, hamburger, notifications, rainbowwallet. rainbowwallet can be themed too, but partially iir.

murderteeth (dismissed):
approved

small things to follow up on

- on mobile the vscrollbar has a lot of margin. this creates dead space around the list that looks like it should scroll when dragged, but doesn't

Image

- button styles behaviors still not consistent. group buttons use a hover transition (easing), the icon and the launcher buttons do not. luancher buttons have a distinct mousedown (active) state, the group and icon button do not.

PR #795, opened 2025-10-16
Chore: update rewards tab and messages

rossgalloway:
### Rewards Tab Display

This PR modifies the behavior of the RewardsTab component that shows staking information on vault pages. The rewards tab is now shown conditionally with the following logic:

Show Rewards Tab if:
currentVault.staking.available=true # staking contract exists
AND one of:
currentVault.staking.rewards.isFinished=false   #rewards are active 
OR
user has staked deposit
OR
user has rewards to claim

Additionally, if rewards are finished, the stake input is hidden.

~~As of writing this PR, the veYFI gauges are not quite fully retired (3 more hours - timestamp 1761782400). At that point, vaults that had gauges should have their deposit and stake behaviour removed and the staking input will be hidden.~~
Gauge staking contracts will always have a running timestamp so we can't use this. Instead I added all the gauges to the blacklist.

### Messaging

The veYFI gauge message that used to show info about liquid lockers has been replaced with a message about the veYFI gauge program ending.

Some other small quality of life items. 
- Unstake no longer claims if set to max. This is confusing and inconsistent. And the button text (exit and claim) was worse. It sounded like you got the underlying back, but it only unstaked. 
- Fixed a decimals bug with vaults that have different decimals than their staking contracts displaying wrong in the table.

### To Test
Run locally and connect wallet. If you navigate to a vault that used to have veYFI staking and you have assets staked or rewards to claim then you should see the rewards tab. If you do not have anything staked or available to claim then you should not. This should be consistent across all staking contracts (veYFI, v3, Arbitrum, etc)

## Related Issue

none

## Motivation and Context

veYFI gauges are deprecaded

## How Has This Been Tested?

locally

image

### DIFF
--- a/apps/vaults-v3/components/details/VaultDetailsHeader.tsx
+++ b/apps/vaults-v3/components/details/VaultDetailsHeader.tsx
@@ -197,87 +197,6 @@ function VaultAPY({
     )
   }
 
-  // if (isSourceVeYFI) {
-  //   const sumOfRewardsAPY = apr.extra.stakingRewardsAPR + apr.extra.gammaRewardAPR
-  //   const veYFIRange = [apr.extra.stakingRewardsAPR / 10 + apr.extra.gammaRewardAPR, sumOfRewardsAPY] as [
-  //     number,
-  //     number
-  //   ]
-  //   const estAPYRange = [veYFIRange[0] + apr.forwardAPR.netAPR, veYFIRange[1] + apr.forwardAPR.netAPR] as [
-  //     number,
-  //     number
-  //   ]
-  //   return (
-  //     <VaultHeaderLineItem
-  //       label={'Historical APY'}
-  //       legend={
-  //         <span className={'tooltip'}>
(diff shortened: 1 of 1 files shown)


murderteeth (approved):

0xeye (dismissed):
LGTM

murderteeth (changes requested):
All my gauged deposits are unwound, and i confirm i no longer see the rewards tab or other info about gauges. We should clean up the APY display tho, most vaults will show 0 rewards now. 
Image

murderteeth (approved):

apps/vaults-v3/components/details/VaultDetailsHeader.tsx
```diff
@@ -735,6 +735,8 @@ export function VaultDetailsHeader({ currentVault }: { currentVault: TYDaemonVau
     }
   }, [blockNumber, refetch, currentVault.chainID])
 
+  console.log(currentVault.staking.source)
```
0xeye: Remove logs and commented code (?)

PR #803, opened 2025-10-22
Add Spectra and Steer bonus messages on v3 VaultActions page

engn33r:
## Description

Add text and links to additional points and yield opportunities on Katana to v3 vaults page.

## Related Issue

Modified in sync with https://github.com/yearn/yearnX/pull/50

## Motivation and Context

Yearn TVL could shift towards Spectra and give users more KAT token yield. But there is an information gap because most users are unaware of this opportunity, so a link is needed to guide users.

## How Has This Been Tested?

Just text and a hyperlink. The text appears and the hyperlink works with `bun run dev`

image


rossgalloway (approved):

murderteeth (approved):
lgtm. we should improve how are tooltips work. you're right it's out of scope here.

ideally tooltip open behavior depends on cursor position in the page. 

for example, here our katana tooltip underflows the bottom of the page if you're opening the tooltip near the bottom of the page.
Image

for an example of opening the tooltip responsively, go here https://seafood.yearn.watch/risk and mouse over the boxes in the risk heatmap.

Image

Image

Image

PR #810, opened 2025-10-30
fix: Spectra Tooltips

rossgalloway:
## Description

* update tooltip to allow mouse navigation to tooltip and links
* adjust styling
* update spectra link to go straight to katana vaults

## Related Issue

https://github.com/yearn/yearn.fi/pull/803

## Motivation and Context

Make it better

image

## How Has This Been Tested?

locally


engn33r:
lgtm, the tooltip now stays open so users can easily click the link in the tooltip text

PR #812, opened 2025-10-31
fix: incorrect import

rossgalloway:
## Description

old `react-router-dom` import was being used and didn't get flagged by linter or build steps. changed to use `react-router`

## Related Issue

failing vercel builds

## Motivation and Context

failing vercel builds

## How Has This Been Tested?

locally

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,6 +1,6 @@
 set -e
 
-echo 'husky: lint error found'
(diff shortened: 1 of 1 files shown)


murderteeth (approved):
approved. but there's a bug on the katana apr where the icons fade out. may have been introduced in previous pr, not sure. will tg a clip

PR #813, opened 2025-11-03
Redirect /_next/image requests to reduce edge usage

murderteeth:
Adds permanent redirect for /_next/image/* paths to localhost to prevent bots from continuing to request these outdated URLs after migration to Vite. This addresses the same edge request optimization issue as f4045e86 and 8127c92c by causing bots to cache the 308 redirect and stop hitting our edge network entirely. 

To test, use the site as usual, should only effect bots.

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -2,6 +2,13 @@
   "installCommand": "bun install --frozen-lockfile",
   "buildCommand": "bun run build",
   "outputDirectory": "dist",
+  "redirects": [
+    {
+      "source": "/_next/image/:path*",
+      "destination": "https://localhost",
+      "permanent": true
+    }
+  ],
   "rewrites": [
     {
       "source": "/proxy/plausible/:path*",


w84april:
Could  you pls help me understand why bots continue requesting /_next/image/* ? 

fwiw tested `curl -I https://yearnfi-git-next-image-yearn.vercel.app/_next/image/test.jpg` and `curl -I https://yearn.fi/_next/image/test.jpg` and got same 403 error. Is this expected or I should do smth else to test?

murderteeth:
> Could you pls help me understand why bots continue requesting /_next/image/* ?
> 
> fwiw tested `curl -I https://yearnfi-git-next-image-yearn.vercel.app/_next/image/test.jpg` and `curl -I https://yearn.fi/_next/image/test.jpg` and got same 403 error. Is this expected or I should do smth else to test?

I think we're getting hit by bots using a headless browser like puppeteer and that's still using a old, cached versions of the app. Hard to say, but the traffic pattern is clear: we're getting blasted with requests for `/_next/image/*` that aren't coming from normal traffic.

PR #827, opened 2025-11-13
Feat: improve risk score tab

rossgalloway:
## Description

This pull request updates the risk scores tab for V3 vaults. 
- The descriptions for each score element for vaults is hidden behind an accordion so it is less cluttered.
- Vault scores are shown and are better described
- styling is updated to make it easier to read
- the RiskScoreIcon from the list is re-used for better understanding.
- Vaults and Strategies are better distinguished 

## Related Issue

none

## Motivation and Context

Bugs reported and also just to make it better

## How Has This Been Tested?

locally. Go to a vault page and click on the risk tab. Then go to a strategy page and click on the risk tab.


w84april (approved):
All looks good but I'd also add some slight animation in accordion to follow existing practice (e.g strategies tab)

PR #828, opened 2025-11-14
Fix: wallet issues

rossgalloway:
## Description

fixes an issue where the wallet seems to have trouble connecting and also showed Invalid Network

useWeb3.tsx
- Removed the useIsMounted() guard from isActive. Wagmi already hydrates persisted sessions safely, so waiting for React’s “mounted” flag was unnecessary and caused false negatives right after refresh.
- isActive now evaluates to isConnected && supportedChain, explicitly allowing chain id 1337. This means as soon as wagmi reports an address and a supported network, our context treats the session as active—even before the second render.

Header.tsx
- Added a new "Connecting..." message that shows on the isConnecting State instead of the "Invalid Network" message

## Related Issue

none

## Motivation and Context

make shit better

## How Has This Been Tested?

locally.

To test, connect your wallet and see what the states show when doing so.

### DIFF
--- a/apps/lib/components/Header.tsx
+++ b/apps/lib/components/Header.tsx
@@ -39,15 +39,24 @@ function Navbar({ nav, currentPathName }: TNavbar): ReactElement {
 function WalletSelector(): ReactElement {
   const { openAccountModal } = useAccountModal()
   const { openChainModal } = useChainModal()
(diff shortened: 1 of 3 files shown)


murderteeth (approved):
nice lgtm

PR #837, opened 2025-12-03
chore: style update

w84april:
## Description

Some style ideas

1. Add 3rd option for vaults header - minimalistic
2. Add carousel in Trending 
3. Mobile adjustments
4. Small stuff (rounded corners etc)


w84april:
> I wonder if we need to show more than 4 suggested vaults

Even if we don't, i thinks it's better to display a row here no matter what the screen size is. Takes too much space otherwise
image

rossgalloway (approved):

PR #843, opened 2025-12-09
fix: Override ox package version to resolve build error.

rossgalloway:
## Description

Current Main branch state is not currently building to prod due to an issue with mismatching versions of the `ox` dependency used by viem, coinbase libs, and porto. This fix sets the version of that sub-dependency across all dependencies using the `overrides` and `resolutions` flags in package.json.

Alternate fix is to revert the viem upgrade to the prior version.

## Related Issue

- Failing build on vercel: https://vercel.com/yearn/yearnfi/GiYmSrt3JtfKsZxRnG5a7eqEBFSu
- PR that broke things: https://github.com/yearn/yearn.fi/pull/841

## Motivation and Context

Fix the site.

## How Has This Been Tested?

locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,9 +60,6 @@
     "vite-plugin-webfont-dl": "3.11.1",
     "vitest": "1.6.1"
   },
-  "resolutions": {
-    "ox": "0.9.6"
-  },
   "overrides": {
     "ox": "0.9.6"
   }


murderteeth (changes requested):
`overrides` is good. do we need `resolutions` too?

PR #847, opened 2025-12-13
Chore: vault page header update

rossgalloway:
## Description

- Updates the Vault page header to use the collapsable header. 
- Clips and adds hover on Vault names that are too long when collapsed mode is active
- updates styling
- vault metadata to chips
- Improves TokenLogo component

Still to do:
- section selector should change as page is scrolled.
- Figure out why things are weird when running dev mode.

## Related Issue

https://linear.app/yearn-webops/issue/FE-81/determine-correct-header-scroll-behavior

## Motivation and Context

Make it better

## How Has This Been Tested?

locally.


w84april (approved):
nice

PR #860, opened 2025-12-17
chore: update katana rewards messages per their changes

rossgalloway:
## Description

Requested Changes from katana:

2 changes to the APY breakdown notes worth updating before tomorrow to avoid user confusion:
- Native APY: We’ve removed the L1 APY. Previously this was double-counted since VB yield was donated directly at the Yearn vault level and already reflected in the Yearn APY’s PPS increase. In v1, Native APY now simply reflects the 30-day Yearn APY.
We’ve also removed the KAT top-up disclaimer. Take a look at the v1 app for reference.
- Base rewards: no change
- App rewards: no change
- Early deposit bonus: no change (this ended back on Sept 1, but we’ve left it displayed for now)

I updated the mouseovers to reflect these changes.

## Related Issue

katana-apr.yearn.fi update coming shortly

## Motivation and Context

Request from integration team

## How Has This Been Tested?

locally

image


apps/lib/hooks/useKatanaAprs.ts
```diff
@@ -2,7 +2,7 @@ import axios from 'axios'
 import { useEffect, useState } from 'react'
 
 const CACHE_KEY = 'katana-aprs-cache'
-const CACHE_TTL = 5 * 60 * 1000 // 5 min
+const CACHE_TTL = 1 * 60 * 1000 // 1 min
```
murderteeth: this could be 15 minutes because the latest cache update sets the apr service's cache to 15 min https://github.com/yearn/katana-apr-service/blob/e097e26f812b03a7ce5b75157211fa76fd6e9deb/src/app/api/vaults/route.ts#L20

murderteeth (approved):
approved with comment

PR #963, opened 2026-01-16
add chip tooltips and improve vault info section

rossgalloway:
## Description:
- This reshapes the expanded section of the Vault List elements to move the Vault info to a place of higher prominence over the charts and makes it persistent. It does this by moving the info to the left side of the expanded section and keeping the strategies and charts on the right. They are toggled and only change on the right side.
- The Vault info section has been re-designed to have info on the chain, type, underling assets, etc with expandable section to get more information.
- Charts in the expanded vault list elements are forced to the 1-year timeframe due to a lack of room for a timeframe selector. Different timeframes are still available in the main vault page.
- New tooltips were added for the vault selector and chain selectors, and all chips now have tooltips
- There is a new fees chip
- The Vault Descriptions has section on the Vaults page has been renamed to "Vault Info" and has the same information as in the expanded vault list section.
- Updated the chart styling to be more "full bleed" and move the Y axis labels inside the chart
- **Removes the trending vaults section for now.**
- add ability to set min TVL in filters menu
- add ability to filter by asset in filters menu

## Screenshots

Screenshot 2026-01-18 at 7 11 39 PM

image

### DIFF
--- a/apps/lib/components/MultiSelectDropdown.tsx
+++ b/apps/lib/components/MultiSelectDropdown.tsx
@@ -59,10 +59,7 @@ function Option(option: TMultiSelectOptionProps): ReactElement {
               {option.icon}
             </div>
           ) : null}
-          <p className={`${option.icon ? 'pl-2' : 'pl-0'} font-normal text-neutral-900`}>
-            {option.label}{' '}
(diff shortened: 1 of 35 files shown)


rossgalloway:
includes changes in #968. Plus removes the trending section (after cleaning up the cards)

0xeye:
Unsure what happened with branch, squashed and rebased it on mine

PR #965, opened 2026-01-17
chore: attempt to refactor resize

w84april:
Attempt to use css instead of resize observer


rossgalloway:
looks like you removed optimism as a supported chain in the selector for the LP token vaults. Can you add that back

PR #1028, opened 2026-02-02
Fix/holdings total balance

w84april:
## Description

- Added reverse lookup map (stakingAddress → vaultAddress) in useWallet.tsx. Workaround to calculate vault balance + staking balance without staking token being actually in a vault list (`staking` only a part of vault now - https://github.com/yearn/kong/pull/332) 
- Staking-only positions now correctly counted in portfolio totals
- Added staking field to Kong schema and mapped it in useFetchYearnVaults.ts
- Updated useSortVaults to use getToken().value for pre-calculated USD values
- Removed manual balance × price calculation

## How to test
Compare your portfolio in yearn.fi and here, should be close


rossgalloway (approved):
LGTM

PR #1051, opened 2026-02-04
chore: cleanup

0xeye:
- Pulls logic reused multiple out into its own component. This way it is easier to maintain consistent layouts
- Combine useFilter logic
- Removes unused files and unused dependencies

To test:
- Run through app, check all features work correctly.


0xeye:
Tested, seems all okay to me

murderteeth (approved):
lgtm. but lets get two approves on this one

PR #1054, opened 2026-02-06
Add cached Curve pool lookup for Curve factory vault LP links

rossgalloway:
## Summary

  - Replaced meta.sourceURI dependency with a Curve API lookup for Curve factory vaults.
  - Added an in‑memory cache (30‑minute TTL) with in‑flight de‑duplication to avoid repeated getPools/all fetches.
  - Uses the LP token address to find the pool and link to poolURLs.deposit[0] in “More Info”.

 ## Key Changes

  - src/components/pages/vaults/components/detail/VaultInfoSection.tsx
      - Added cache + helper to fetch Curve pools.
      - Render Curve link from Curve API data for factory Curve vaults.

 ## How to Test

  1. Open a Curve factory vault page.
  2. In “More Info”, confirm the Curve link appears and points to the pool’s deposit URL.
  3. Navigate between Curve factory vaults and verify the link still resolves without repeated network fetches (cache
     hit).

### DIFF
--- a/src/components/shared/utils/curveUrlUtils.test.ts
+++ b/src/components/shared/utils/curveUrlUtils.test.ts
@@ -17,6 +17,11 @@ describe('normalizeCurveUrl', () => {
   it('returns empty string for invalid urls', () => {
     expect(normalizeCurveUrl('not-a-url')).toBe('')
   })
+
+  it('returns empty string for non-http schemes', () => {
+    expect(normalizeCurveUrl('javascript://curve.fi/%0Aalert(1)')).toBe('')
+    expect(normalizeCurveUrl('ftp://curve.fi/pools')).toBe('')
+  })
 })
 
 describe('isCurveHostUrl', () => {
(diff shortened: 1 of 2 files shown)


murderteeth (changes requested):
here would it be better to use a tan query with staleTime set instead of the useEffect + naked fetch + custom ttl implementation?

also should all curve pools have links? this one does not for example https://yearnfi-git-add-curve-link-yearn.vercel.app/vaults/1/0x241AdD131B9aaa7527132b752252b99420937ADc

murderteeth (approved):
## ✅ Approved

### Note: Category-based scope limitation

Some Curve LP vaults will still be missing links due to the `isCurveCategory` gate. Example: [`0x241AdD131B9aaa7527132b752252b99420937ADc`](https://yearnfi-git-add-curve-link-yearn.vercel.app/vaults/1/0x241AdD131B9aaa7527132b752252b99420937ADc) has `category: "Stablecoin"` (not `"Curve"`), `sourceURL: null`, but its LP token exists in the Curve API with a valid deposit URL.

PR #1067, opened 2026-02-16
feat: init portfolio history

w84april:
## Description

This branch expands the portfolio backend from a historical balance chart API into a broader holdings platform with:

- a cached holdings history API
- a FIFO-based portfolio PnL API
- a vault-level drilldown API for accounting inspection
- better handling for staking, migrations, router flows, rewards, and selected CoW settlement buys
- more reliable price fetching and caching
- frontend integration for Portfolio PnL, Economic Gain, and stable vs volatile breakdowns

It also adds operational support for lazy cache invalidation when new vaults are indexed, plus cleanup chores for stale cache rows.

### Public API surfaces

- `GET /api/holdings/history`
  - returns chart-ready daily USD portfolio history
  - uses cache-first aggregation and recalculates only missing days
  - supports `version`, `refresh`, `fetchType`, and `paginationMode`
- `GET /api/holdings/pnl`
  - returns compact portfolio PnL summary plus vault-family rows
  - supports `strict`, `zero_basis`, and `windfall` handling for unknown-basis transfer-ins
  - exposes both `totalPnlUsd` and `totalEconomicGainUsd`
  - includes stable vs volatile category breakdowns
- `GET /api/holdings/pnl/drilldown`
  - returns the same valuation basis as `/pnl`, but expanded with lots, realized consumption, unknown-basis receipts/withdrawals, and journal-style inspection data
  - intended for detailed accounting/debug UI
- `POST /api/admin/invalidate-cache`
  - records vault invalidations so users holding newly indexed vaults refresh lazily on next request
- `POST /api/holdings/chores`
  - cleanup endpoint for stale cache maintenance

## Accounting Model

The PnL engine now models Yearn positions as vault-family FIFO lots rather than simple balance snapshots.

It:
- groups underlying vaults and staking wrappers into a single family
- tracks lot location (`vault` vs `staked`) instead of treating staking as a new investment
- uses acquisition-time PPS and acquisition-time token prices for known-basis USD accounting
- separates `Portfolio PnL` from `Economic Gain`
- supports three unknown-basis policies:
  - `strict`: exclude ambiguous lots from PnL
  - `zero_basis`: treat ambiguous receipts as zero-cost basis
  - `windfall`: split receipt-time value from later market PnL
- keeps conservative partial-basis behavior when full reconstruction is not possible

### History flow

1. Load cached daily totals.
2. Detect stale cache via vault invalidation timestamps.
3. Fetch Envio events when missing days require recomputation.
4. Build a share-balance timeline.
5. Resolve vault metadata and PPS from Kong.
6. Fetch underlying asset prices from DefiLlama.
7. Compute daily USD totals.
8. Persist results back to cache.

### PnL flow

1. Fetch address-scoped events.
2. Enrich with tx-scoped context for same-hash router, staking, and migration flows.
3. Normalize into vault-family ledgers.
4. Build FIFO lots and realized entries.
5. Value remaining lots with current PPS and token price.
6. Apply the chosen unknown-basis policy.
7. Materialize summary rows and drilldown payloads.

## Data Sources

The holdings and PnL pipeline now combines:

- Envio GraphQL for deposits, withdrawals, and transfers
- Kong for vault metadata and historical/current PPS
- DefiLlama for historical/current USD prices of the vault asset token
- Ethereum RPC + CoW settlement logs for selected known-basis synthetic acquisition enrichment

## Pricing / Performance Improvements

This branch also hardens price fetching significantly:

- tuned DefiLlama batching for reliability
- pro/free GET fallback behavior
- request splitting on infra-sized failures
- exact positive-price caching
- exact negative-cache for unsupported price points with TTL
- support for `fetchType=parallel` and `paginationMode=all` on both history and PnL surfaces
- support for the same fetch controls on Vercel handlers, not only the Bun server

## Frontend Integration

The Portfolio page now consumes the new APIs and surfaces:

- chart history from `/api/holdings/history`
- Portfolio PnL and Economic Gain from `/api/holdings/pnl`
- tooltip copy explaining the metric formulas
- stable vs volatile breakdowns under PnL and Economic Gain
- updated response typing and schema validation

## Operational Changes

This branch adds lazy cache invalidation for newly indexed vaults:

1. After deploying an indexer update with new vaults, call `POST /api/admin/invalidate-cache`.
2. Vault invalidation timestamps are recorded in the database.
3. On the next user request, the system checks whether any held vaults were invalidated after cache creation.
4. If stale, the affected user cache is cleared and recalculated lazily.

It also adds a chores endpoint for stale cache cleanup.

## Documentation

This branch ships with dedicated docs for both the high-level holdings system and the accounting model:

- `api/lib/holdings/README.md`
- `api/lib/holdings/PNL.md`

## Testing

The work includes substantial coverage across:

- history aggregation
- GraphQL event fetching
- DefiLlama batching and caching
- Kong metadata/PPS integration
- CoW settlement enrichment
- PnL mode behavior
- vault metadata resolution
- drilldown and summary materialization

## Screenshots
image


api/lib/holdings/README.md
```diff
@@ -0,0 +1,260 @@
...
+    ...
+  ]
+}
+```
+
+### DELETE `/api/v1/holdings/cache`
+Clear cached data (useful for forcing recalculation).
+
+```bash
+curl -X DELETE "http://localhost:3001/api/v1/holdings/cache?address=0x..."
```
0xeye: Is this permissioned access? We shouldn't have DELETE api endpoints exposed publicly

Also, can we merge in [feat/portfolio-history-adjustments](https://github.com/yearn/yearn.fi/tree/feat/portfolio-history-adjustments) - it just replaces let/for to const/maps, something I've expanded the CLAUDE.md doc to include in future

0xeye (changes requested):

api/lib/holdings/README.md
w84april: Oh! Sure, this was added for simpler testing purposes only. Thanks for pointing it out

murderteeth (changes requested):
Two changes requested:

**1. Rate limit `/api/holdings/history`**

Every uncached request hits Envio 6 times and writes to the database if the address has holdings. Add a `middleware.ts` that limits this endpoint by IP (~10 req/min), returning 429 when exceeded.

**2. Daily cleanup for stale cache rows**

Cached rows outside the 365-day query window are never deleted and accumulate indefinitely. Add `api/holdings/chores.ts` that deletes them:
```sql
DELETE FROM holdings_totals WHERE date < NOW() - INTERVAL '366 days'
```

Trigger via Vercel cron in `vercel.json`:
```json
{ "crons": [{ "path": "/api/holdings/chores", "schedule": "0 3 * * *" }] }
```

Gate with `CRON_SECRET` check.

murderteeth (changes requested):
## Review

Nice work on the architecture — clean separation between GraphQL fetching, position timeline, price resolution, caching, and aggregation. SQL is parameterized throughout, cron cleanup is properly secured, and the frontend chart component is solid.

A few things to address before merge:

### Security

**1. Error response leaks internals** (`history.ts:149-154`)

The catch block returns `message`, `envioUrl` status, and `authEnabled` to clients. In production this should be a generic error — internal diagnostics shouldn't be exposed to callers.

**2. Debug logging should be removed** (`history.ts:35-36, 45, 64, 67, 72, 76, 100-101`)

IP addresses, DATABASE_URL presence, and rate limit internals are logged on every request. There's a commit specifically adding these (`be738aad debug: add logs`). Please remove before merge.

**3. Stack trace in error response** (`api/server.ts` ~line 203)

`error.stack` is returned in the response body from `handleHoldingsHistory`. Remove this — if you need to inspect errors during dev, use browser devtools, server logs, or a debugger rather than returning stack traces over HTTP.

**4. Remove debug and breakdown endpoints** (`api/server.ts`)

`/api/holdings/debug` and `/api/holdings/breakdown` expose detailed internal state (raw events, vault metadata, PPS values). Remove these regardless of auth — they shouldn't ship to production. During dev, tools like the Network tab, server-side logging, or test scripts are better ways to inspect this data.

**5. `chores.ts` leaks config state** (`chores.ts:11`)

The `"CRON_SECRET not configured"` message tells callers about internal configuration. Still log it server-side (`console.error`), but return a generic 500 to the client.

### Code duplication

**6. Duplicated rate limiting**

`history.ts` has its own inline `checkRateLimit` (lines 33-82) while `api/lib/holdings/services/ratelimit.ts` has a proper service version exported from `index.ts`. The service version is never used. Consolidate to one implementation.

### Minor

**7. Unused Zod schemas in `types/api.ts`**

The detailed schemas (`chains`, `vaults` breakdown) are defined but only the simple response is used on the client. Dead code?

murderteeth (changes requested):
## Review

Previous feedback looks addressed — debug logs removed, debug/breakdown endpoints removed, error responses gated behind `NODE_ENV`, rate limiting consolidated, config state no longer leaked from chores, unused Zod schemas cleaned up. 👍

A few things remaining:

### 1. Why 2-minute timeout on `baseFetcher`? (`fetchers.ts:90`)

This changes the timeout from 15s to 120s for _all_ callers of `baseFetcher`, not just the holdings endpoint. Why do we need this? If it's only for holdings, consider passing timeout as a parameter or creating a dedicated fetcher so we don't mask slow responses across the rest of the app.

### 2. New DB connection per rate limit check (`ratelimit.ts:18`)

Creates a fresh `neon()` connection on every call instead of reusing the pool from `db/connection.ts`. Inconsistent with the rest of the holdings services and less efficient.

### 3. Typo (`server.ts:24`)

`handleCorsPrelight` → `handleCorsPreFlight`

rossgalloway:
image

if we have an error and don't return anything, this should probably collapse or be hidden and not take up space.

murderteeth:
> image
> if we have an error and don't return anything, this should probably collapse or be hidden and not take up space.

for the empty state (not the same as error state) instead of collapse or hide, lets show a placeholder\mockup. something that communicates "this is how awesome your portfolio would look if you had one" followed with a clear call to action, Explore Vaults, or whatever. use muted colors and\or a badge that clearly says "example", so it's clear we're not showing real holdings.

the portfolio page is popular, users spend a lot of time on it. we dont want to undersell it.

w84april:
@rossgalloway fyi closing in favour of #1201

PR #1150, opened 2026-03-31
feat: reorg token list

0xeye:
Reprioritises token rows to prioritize known assets

Before: 
image

After:
image

image


rossgalloway:
This cleans up the main list, but I get a bunch of legit tokens in the "unlisted" section, and they arent sorted by USD value anymore

The behavior is better for me in the main branch. I think the fixes that were incorporated in the yvUSD withdrawal update need to be incorporated into the release branch. That is what some of the merge conflicts are

PR #1151, opened 2026-04-01
Fix deposit success balance refresh lag

rossgalloway:
## Summary

- Start the post-deposit wallet balance refresh when the transaction is confirmed on-chain, **before** the success screen and confetti appear, rather than waiting until after confetti ends
- Add investigation notes for the follow-up context re-render optimization work

## Problem

With `deferSuccessEffectsUntilConfettiEnd=true` (the default in `WidgetDeposit`), `onAllComplete` fires only after the confetti animation ends — roughly 2–3 seconds after the transaction confirms. The async RPC balance fetch therefore started late, and the resulting re-renders hit the user just as they were interacting with the UI after dismissing the overlay.

## What changed

**`src/components/pages/vaults/components/widget/deposit/index.tsx`**

Split the post-deposit work into two callbacks:

| Callback | Trigger | Does |
|---|---|---|
| `handleDepositTransactionSuccess(label)` | `onStepSuccess` — fires when tx confirms, before success screen | `refreshWalletBalances` + `refetchVaultUserData` |
| `handleDepositSuccess` | `onAllComplete` — fires after confetti ends | Analytics tracking, input reset, `onDepositSuccess?.()` |

For cross-chain deposits the `TransactionOverlay` does not emit `onStepSuccess`, so the balance refresh remains in `onAllComplete` there (source-chain tokens change immediately so refreshing them is still correct).

## Test plan

- [ ] Deposit into a vault and verify the balance in the vault row updates by the time the success overlay is dismissed (no lag after close)
- [ ] Cross-chain deposit still triggers balance refresh after the success overlay closes
- [ ] Approval step does not trigger an early balance refresh
- [ ] Stake flow (DIRECT_STAKE route) early refresh fires correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rossgalloway:
superseded by #1155

PR #1152, opened 2026-04-01
release rebase

0xeye:
we do a little rebase


0xeye:
for diff check purposes, can revert here

PR #1154, opened 2026-04-01
chore: bulk remove tests

0xeye:
Tests guard math where a wrong decimal can silently lose user funds — not UI or structural code. When AI writes both the implementation and the tests, it validates its own assumptions against itself. When the implementation changes, the tests get rewritten to match, so they never actually catch anything.

Update CLAUDE.md to factor this in

```
KEEP:
  1. strategiesPercentFormat.test.ts — percent rounding rules
  2. format.test.ts — locale-aware currency formatting
  3. vaultApy.test.ts — APY/APR compound interest math
  4. useVaultApyData.test.ts — Katana APR computation
  5. valuation.test.ts — deposit value, price impact %
  6. cooldownUtils.test.ts — duration/time math
  7. yvUsd.test.ts — weighted APY, PPS conversions, APR annualization, bigint math
  8. useYvUsdCharts.helpers.test.ts — PPS series, APY derivation
  9. YvUsdWithdraw.helpers.test.ts — share/asset bigint conversions
  10. useYvUsdVaults.helpers.test.ts — TVL breakdown math
  11. kongVaultSelectors.test.ts — pricePerShare conversion, APY selection
  12. useVaultFilterUtils.test.ts — USD valuation from pricePerShare```


0xeye:
Going to think on this some more

PR #1158, opened 2026-04-03
Add high slippage protection

rossgalloway:
## Summary
  - replace the old split slippage model with a single user tolerance that covers both quote impact and Enso execution slippage
  - rework deposit and withdraw zap flows to use Enso `minAmountOut` for worst-case gating and for the withdraw `You will receive at least` display
  - harden Enso quote handling for zero-output routes, missing `tx.chainId`, 4xx/422 error payloads, stale quotes, and slippage-setting changes
  - require typed acknowledgement above 1% slippage, hard-block zap flows at 5%, and fix withdraw zap USD displays
  - keep the local API server configurable via env while preserving `3001` as the default in code

  ## Why
  The original branch only added a frontend price-impact warning/block, but the user-configured zap slippage and the frontend quote-quality checks were still separate. That meant a
  user could set a low slippage tolerance and still receive routes whose total worst-case outcome exceeded that value.

  There were also several Enso-specific edge cases that made the flow unreliable in practice:
  - zero-output withdraw routes were not surfaced early enough
  - some valid Enso success payloads omitted `tx.chainId`
  - some Enso failures returned `message`/`description` payloads instead of a standard `error`
  - changing slippage could briefly leave stale quotes and flickering red warnings in the UI
  - withdraw zap USD displays were derived from compact strings instead of raw quote values

  ## Impact
  - zap flows now treat the saved slippage setting as the maximum total acceptable loss, not just Enso execution drift
  - deposit/withdraw quotes are evaluated against worst-case output using `minAmountOut`
  - if worst-case total slippage exceeds the saved tolerance, the CTA stays disabled and the user is directed to raise tolerance in settings
  - slippage above 1% requires typing `I accept the risk that I may lose money doing this`
  - zap flows at or above 5% total slippage are blocked entirely
  - withdraw zap UI now distinguishes between:
    - quoted output shown in the zap preview
    - minimum received shown in `You will receive at least`
  - quote changes now clear stale Enso results immediately and show loading instead of flashing temporary red “bad quote” states

  ## Validation
  - `bun run lint:fix`
  - `bun run tslint`
  - `bunx vitest run src/components/shared/utils/slippage.test.ts src/components/pages/vaults/hooks/solvers/ensoRoute.test.ts src/components/pages/vaults/components/widget/deposit/
  valuation.test.ts src/components/pages/vaults/components/widget/withdraw/valuation.test.ts`

To Test:

- Impersonate the treasury contract in your wallet: 0x93A62dA5a14C80f265DAbC077fCEE437B1a0Efde
- select any vault with large holdings (USDC-1 is great)
- try to withdraw via zap to WBTC with a largish sum. test different values to see different slippages.
- change slippage tolerances in the widget settings.
- confirm that you need to type `I accept the risk that I may lose money doing this` to set slippage over 1%.
- confirm that you cannot transact with slippage over 5%


rossgalloway:
I'm still not super happy with how the slippage and extra widget info is being displayed, but it is getting better.

w84april (approved):

PR #1160, opened 2026-04-06
fix: correct withdraw zap usd displays

rossgalloway:
## Summary
- fix withdraw zap USD display to use raw quote amounts instead of compact display text like `16.1K`
- add missing USD values to the withdraw zap detail rows for `You will swap` and `You will receive at least`
- keep non-zap withdraw behavior unchanged and add regression coverage for both issues

## Problem
The withdraw zap flow had two related UI bugs:
- the output USD value under the selected zap token was being derived from formatted display text, which caused mis-scaled values like `$17.85` instead of roughly `$17,850`
- the withdraw details section showed token amounts for zap rows without their USD equivalents

## Implementation
- extend the withdraw zap token metadata with raw amount + decimals so the UI can compute USD values from quote data
- update `InputTokenAmount` to format zap output USD from raw quote values via shared formatting helpers instead of `parseFloat` on compact text
- update `WithdrawDetails` to append USD values for the zap swap/receive rows only
- add focused tests for the compact-value parsing regression and the missing USD detail rows

## Verification
When zapping out of a position in a withdrawal transaction, all units should be correctly shown and the more info section below the input box should show USD values as well as unit values

### DIFF
--- a/plans/2026-04-06-withdraw-zap-usd-amounts.md
+++ b/plans/2026-04-06-withdraw-zap-usd-amounts.md
@@ -1,0 +1,438 @@
+# Withdraw Zap USD Amounts Implementation Plan
+
+> **For Hermes:** Use subagent-driven-development skill to implement this plan task-by-task.
+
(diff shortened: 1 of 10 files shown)


rossgalloway:
I folded these changes into #1158.

PR #1169, opened 2026-04-08
fix: make production builds memory-safe

rossgalloway:
## Summary
- make production sourcemaps opt-in instead of always enabled
- keep the `/icon-list` audit route dev-only so it is not bundled into production
- narrow the icon-list raw source glob to the current Vite syntax and exclude test/spec files

## Why
`bun run build` was being OOM-killed after module transforms during Rollup chunk rendering. The immediate trigger was production sourcemap generation on a large wallet-heavy graph. The icon audit page was also pulling the entire `src` tree, including test files, into the production bundle via a raw `import.meta.glob`, which made the graph larger than it needed to be.

## Verification
- `bun run build`
- `bun run tslint`
- `bun run lint:fix`
- confirmed `dist/assets` no longer contains `*.test.*` or `*.spec.*` chunks

## Notes
- use `bun run build:sourcemap` when a sourcemap build is explicitly needed
- the Rollup `/*#__PURE__*/` warnings from vendored `ox` files still appear, but they are non-fatal


murderteeth (changes requested):
## Icon-list changes look good

Making `/icon-list` dev-only and excluding test/spec files from the glob are solid improvements — that page has no business in the prod bundle. The `import.meta.glob` syntax modernization is a nice bonus.

## Sourcemap default should stay as-is

Sourcemaps in production are valuable — when users screenshot console errors, you get readable stack traces instead of minified gibberish. Disabling them globally to work around a local resource constraint is the wrong tradeoff.

**Suggestion:** keep `sourcemap: true` as the default (matching current behavior), and add a `build:no-sourcemap` script for devs who need a lighter build locally:

```json
"build": "tsc && vite build",
"build:no-sourcemap": "tsc && vite build --no-sourcemap"
```

That way prod stays unchanged and you have an escape hatch when your machine is under pressure.

rossgalloway:
> ## Icon-list changes look good
> 
> Making `/icon-list` dev-only and excluding test/spec files from the glob are solid improvements — that page has no business in the prod bundle. The `import.meta.glob` syntax modernization is a nice bonus.
> ## Sourcemap default should stay as-is
> 
> Sourcemaps in production are valuable — when users screenshot console errors, you get readable stack traces instead of minified gibberish. Disabling them globally to work around a local resource constraint is the wrong tradeoff.
> 
> **Suggestion:** keep `sourcemap: true` as the default (matching current behavior), and add a `build:no-sourcemap` script for devs who need a lighter build locally:
> 
> ```json
> "build": "tsc && vite build",
> "build:no-sourcemap": "tsc && vite build --no-sourcemap"
> ```
> 
> That way prod stays unchanged and you have an escape hatch when your machine is under pressure.

changes made

PR #1171, opened 2026-04-09
Create sec.txt

shankyysecurity:


murderteeth:
thank you, we are adding a correct security txt soon

Issue #1164, opened 2026-04-07
Filter Merkl rewards to Yearn products and refresh immediately after claim


rossgalloway:
## Summary
Improve Merkl rewards handling in yearn.fi by filtering to Yearn-product rewards and refreshing state immediately after claim.

## Problems
1. Rewards/claims may include items that are not clearly limited to Yearn-product rewards.
2. Claimed rewards state does not refresh immediately after a successful claim.

## Proposed changes
- filter the Merkl rewards surface to only show relevant Yearn-product rewards (example: KAT rewards)
- refresh rewards state immediately after successful claim completion

## Acceptance criteria
- rewards UI is limited to the intended Yearn-product rewards set
- successful claims update the UI immediately rather than waiting for delayed eventual clearing

## Tracker
- To-do ID: T-013

PR #1180, opened 2026-04-10
Add merkle rewards filters and make them refresh

rossgalloway:
## Description

- This pull request updates the way the site deals with Merkl.xyz rewards. It filters out all rewards that are not earned by our vaults. Essentially, this means it returns only the Katana Rewards. 
- It also updates how they are treated so that claiming them immediately removes them from the list of rewards, unlike the prior behavior where the rewards stayed as claimable.
- There were also some bugs with the claim process on the portfolio page that have been fixed.

## Related Issue

https://github.com/yearn/yearn.fi/issues/1164

## Motivation and Context

Make it better

## How Has This Been Tested?

locally. To test, you need to have deposited into a Katana vault. Rewards should start flowing relatively quickly. Once you have rewards, try claiming from both the vault page and the portfolio page. Claims should be seamless and disappear after claiming.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "dev:ts": "tsc --watch",
     "tenderly": "bun scripts/tenderly.ts",
     "build": "tsc && vite build",
-    "build:no-sourcemap": "tsc && vite build --no-sourcemap",
+    "build:sourcemap": "tsc && vite build --sourcemap",
     "preview": "concurrently \"bun run preview:server\" \"bun run preview:client\"",
     "preview:client": "vite preview",
     "preview:server": "bun scripts/ensure-api-server.js",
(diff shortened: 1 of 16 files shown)


w84april (approved):
Image
Still have no logos for some rewards, but i think all good.
noop but maybe we should just add those assets:
"0x7F1f4b4b29f5058fA32CC7a97141b8D7e5ABDC2d"
"0x3ba1fbC4c3aEA775d335b31fb53778f46FD3a330"

rossgalloway:
> Still have no logos for some rewards, but i think all good. noop but maybe we should just add those assets: "0x7F1f4b4b29f5058fA32CC7a97141b8D7e5ABDC2d" "0x3ba1fbC4c3aEA775d335b31fb53778f46FD3a330"

new PR to add these: https://github.com/yearn/tokenAssets/pull/54

PR #1181, opened 2026-04-10
fix: surface filtered vault search matches

rossgalloway:
## Summary

This PR improves vault search recovery when matches are being filtered out instead of actually missing.

- Adds a bottom-of-list recovery row on `/vaults` when a search still has matching vaults hidden by active filters.
- Reworks blocking-filter insights to group the actual filters hiding those matches and show actionable recovery buttons with `+N` result counts.
- Treats `minTvl=0` as the true disabled state for low-TVL filtering, so the recovery action can fully surface low-TVL matches.
- Removes the legacy `showHiddenVaults` toggle from the filters UI and stops reading or persisting `showHidden` state from query params or local storage.
- Adds focused tests for blocking-filter grouping, low-TVL recovery, and the new search-recovery row.

## Verification

Passed locally:

- `bun run tslint`
- `bun run lint:fix`
- `bunx vitest run src/components/pages/vaults/hooks/useVaultsPageModel.blocking-filters.test.tsx src/components/pages/vaults/index.search-recovery.test.tsx src/components/pages/vaults/utils/blockingFilterInsights.test.ts`
- `bun run build`

Repo-wide baseline note:

- `bun run test` is not green on this branch, but the failing suites are pre-existing and reproduce on `origin/main` as well:
- `src/components/pages/vaults/components/widget/yvUSD/YvUsdWithdraw.helpers.test.ts`
- `src/components/pages/vaults/components/SuggestedVaultCard.test.tsx`
- `src/components/pages/vaults/components/filters/VaultVersionToggle.test.tsx`

## Reviewer Test Plan

1. Run `bun run dev`.
2. Open `http://localhost:3000/vaults`.
3. Search for a vault/token where at least one matching result has under $500 in TVL.
4. Confirm the list shows a bottom recovery row with actions such as `Show low-TVL vaults (+N)` instead of implying the search has no more matches.
5. Click the recovery action and confirm additional matching vaults appear.
6. Repeat with another restrictive filter combination such as chain, category, aggressiveness, or vault type and confirm the recovery row offers the relevant filter-relaxation action(s).
7. Open the advanced filters area and confirm `Show hidden vaults` is no longer present.
8. Visit `/vaults?showHidden=true` and confirm the page ignores that legacy param: no hidden-vault toggle appears, and the URL/state does not preserve `showHidden` after interacting with filters.


w84april (approved):
tested lg

PR #1182, opened 2026-04-10
Devex: improve tenderly and dev tools

rossgalloway:
## Summary
- add a local Tenderly mode toggle in the header with the updated single-badge interaction and safer disable behavior. This means you can keep the `VITE_TENDERLY_MODE` env var true and toggle between tenderly mode and normal mode in dev builds more easily.
- add shared dev/preview API launchers with port selection, portable port detection, and mode-aware env handling. When you run `bun run dev` or `bun run preview`, a small CLI picker pops up and lets you change the port of the API. This is useful if you are running more than one dev server at a time and one of them makes changes to the API server.
- harden API reuse so it only reuses matching local runtime state and only invalidates on API-relevant committed/env changes

## Testing
- bun run lint:fix
- bun run tslint
- bunx vitest run scripts/api-runtime.test.ts scripts/dev.test.ts scripts/preview.test.ts src/config/tenderly.test.ts
- bun run build

## to test
- set  `VITE_TENDERLY_MODE` to true and make sure your tenderly testnets are set up. you should be able to toggle between the testnets and live chain RPCs with the toggle. clicking on the toggle badge still opens the toolbar.
- try running `bun run dev` and `bun run preview` and you should be able to choose your api server port.


w84april (approved):
tested lg

PR #1188, opened 2026-04-15
Feat/pnl simple

w84april:
## Description

Simpler PNL calculation that accounts protocol return only (and approximates USD value generated)

image


rossgalloway:
@w84april ok to close this one?

w84april:
@rossgalloway guess so yeah

PR #1200, opened 2026-04-23
fix clipped search results

rossgalloway:
## Description

Bug report that empty search result filters were getting clipped if there were lots of them.
image

## Motivation and Context

fix it

## How Has This Been Tested?

locally. To search. Select single asset vaults, ethereum, and search for compound. nothing in the resulting prompt to widen your search should be cut off.

image


murderteeth (approved):
approved. VaultsListEmpty.test.tsx feels fragile testing implementation vs behavior

PR #1201, opened 2026-04-24
May Release Branch

w84april:
…190)

* feat: init portfolio history

* fix: build

* fix: getEnv

* fix: env

* fix: localhost connection attempt

* chore: debug

* chore: attempt

* feat: add v2 support

* fix: transfers calculation

* chore: rm clearCache

* feat: calculate gauge

* fix: loading state

* fix: optimize pps fetch

* chore: rename

* feat: add breakdown

* feat: store aggregated data

* chore: upd docs

* feat: add seenTokens

* feat: always recalculate today

* fix: use only 12:00:00 AM timestamps

* feat: add 1y option

* feat: cache prices

* feat: doc

* feat: add batching for queries

* feat: doc

* feat: sequential + batching

* upd: doc

* fix: db spam

* feat: rate limit & clear stale cache

* debug: add logs

* fix: query

* chore: do not expose internals

* chore: remove logging

* chore: misc

* chore: cleanup

* chore: rm dead code

* feat: invalidate cache

* chore: address comments

* chore: remove refresh for safety

* chore: return clear cache for local

* chore: lint

* Feat/portfolio pnl 2 (#1125)

* feat: init pnl

* feat: consider staking

* feat: omit migrate

* chore: support migrate

* feat: 3 calc types

* feat: add totalEconomicGainUsd

* chore: docs

* chore: refactor

* feat: actually consider price at deposit time

* chore: optimize price fetch

* fix: cache

* chore: make fetching more aggressive

* feat: test params

* feat: improve totalCurrentValueUsd

* feat: share events & kong

* feat: fe + history fix

* fix: defillama batching

* feat: tooltips

* feat: handle cow

* feat: tune defillama

* chore: tune a bit more

* feat: add missing prices cache

* feat: docs

* fix: timeout

* chore: copy

* feat: stable vs volatile

* Feat/portfolio pnl excessive (#1148)

* feat: init excessive

* chore: naming

* feat: actually add drilldown

* fix: v2/v3 filter

* fix: version

* fix: env

* feat: cover crvusd rewards

* feat: vault to vault migrate via enso

* feat: init katana rewards

* chore: upd receiver address

* feat: fix bold

* feat: add fetchType + paginationMode

* feat: support fetchtype + pagination on vercel

* feat: hash user address

* feat: ensure schema is initialized

* chore: lint

* fix: katana pricing

* feat: dummy chart

* feat: breakdown

* feat: breakdown for timestamp

* feat: merge yvUSD

* fix: nasty scroll

* chore: lint

* chore: hide if isHidden=true

* feat: make it case insensitive

* feat: init simple

* feat: display

* add annualized return and ETH denominated charting

* Styling and add all time duration

* add charts

* fix staked asset withdrawal issue

* fix issues with staked shares

* fix simple return accounting and growth index charts

* add simple return chart explainers

* add growth chart numeraire spec

* Finalize growth display modes

* Split growth index chart modes

* Keep ETH growth history when partial vaults exist

* remove overall cumulative chart

* improve charts and add to portfolio rows

* Feat/chart tweaks (#1193)

* fix: rescale growth

* fix: prevent flickering

* style: flickering & mobile

* feat: sort overlay by index

* feat: add growth position & index

* fix: growth index empty values

* feat: connect the dots

* feat: consistent chart

* Add position charts to vault pages with balance and to portfolio expanded views. other minor tweaks

* fix review comments

* fix review issues and dark mode tooltip

* feat: dedupe transfer and stake

* feat: improve aggregator (#1195)

* feat: improve aggregator

* chore: lint

* Feat/indexed history (#1198)

* feat: init activity

* feat: enrich

* chore: rm address

---------


rossgalloway:
I will be merging the `prep for yvBTC` branch. It has a commit that disables a cache is the DB is not available. I'm not sure if this is the best behavior, but is not blocking. I made an issue for it here: https://github.com/yearn/yearn.fi/issues/1227

murderteeth (changes requested):
- we should rate limit using vercel's firewall instead of rolling our own. vercel's firewall is built for this and gives us observability. and we should cover all sensitive or potentially expensive endpoints.

- why is caching turned off on enso balances `export const ENSO_BALANCES_CACHE_CONTROL = 'private, no-store, max-age=0, must-revalidate'`

- i see several endpoints that only declare client cache. follow the caching pattern in api/vault/meta.ts for using vercel's edge cache or explain why that's not wanted. for example
```
res.setHeader('Content-Type', 'text/html')
res.setHeader('Vercel-CDN-Cache-Control', 'public, s-maxage=86400, stale-while-revalidate=604800')
res.setHeader('Cache-Control', 'public, max-age=0, must-revalidate')
```

- replace "Admin invalidation: /api/admin/invalidate-cache" with a secure backend process
-- how often will we need this? can we just do it manually from upstash ui

- re prices
-- the code fallsback to defillama silently. if we need the fallback, we should measure how often it's needed. ask matheus for help instrumenting the backend with Sentry
-- the yearn-price api is -compatible- with defillama's api, but that doesn't mean we should call it the same way using the same pagination\batching. there are probably more efficient ways to call our price api, ask matheus for guidance
-- unknown chainid defaults to eth. this feels like an error to me and not a good default. 
--- api/lib/holdings/services/defillama.ts:81
--- api/lib/holdings/services/nestedVaultPrices.ts:23

- re the invalidation key system. it feels over complicated to me. but what sticks out is the invalidation keys have no TTL. shouldn't they have at least a 30d TTL?
-- also cache staleness check fails open. the error is only logged server-side. the request still proceeds as if the cache is known fresh. The handler should either recompute, fail, or explicitly return degraded/staleness-unknown data rather than treating redis check failure as not stale

- no plausible measurements. these features are expensive. we should know how they perform.

- `concurrently` is now an unused dep in package.json. out of scope but related, why was concurrently replaced with custom bootstraping

w84april:
thx @murderteeth great review!
- added vercel rate limiting, killed redis limit
- http caching is turned off for enso balances because we cache it on tanstack/query side. we are able to control that fetch, to re-fetch etc
- did follow the cache pattern here
- killed cache invalidation route. added a guide on how to add that to redis
-  it's not actually falling back to defillama. if some price is missing in yearn-prices we just fallback to 0. the only defillama fallback case - no bearer token in env vars present that is needed for yearn-prices fetch. this is fine imo
- there are definitely better ways but i'd leave all the backend changes to next release. once again been checking what can be improved on FE only regarding price fetch and found some bottleneck (only appeared with 'ALL' filter applied in portfolio). fixed that
- killed default to eth logic
- yeah invalidation logic seems complex. what i'd do is to keep it alive for couple months and kill it if not used
- as long as we manage cache invalidation manually now TTL is to be set there
- added plausible analytics to check the load time for expensive queries
- removed concurrently

murderteeth (approved):
approved, thanks for addressing so many of these. two notes:

"http caching is turned off for enso balances because we cache it on tanstack/query side. we are able to control that fetch, to re-fetch etc" -- yes but with no server side caching the backend api is now griefable. someone could spam the balance end point and rate limit our enso key. that's ok, caching is just one way to gaurd that. try adding it to the exisintg rate limit rule. that's better i think.

"added plausible analytics to check the load time for expensive queries" -- i miscommunicated here mb! when i said "we should know how they perform" i meant "perform" in terms of product analytics. i want to know how engaged users are with the portfolio tool and if\how it leads them to other products on the app. Otoh measuring query time is good too, but we have other methods for that.

netier of these are blocking. but please update the rate limit rule and consider another plausible update asap.

PR #1204, opened 2026-04-28
Feat/history UI

w84april:
## Description

- Display zaps in activity
- Tweak style according to things discussed on UX call


rossgalloway:
sorry, I looked at this a while ago and never left comments. Generally much better than before! A few things I'd like to see:

- Remove the padding between elements and at the right and left sides. Lets make it feel more like the main vault list where the row takes the full space.
- On expand, I don't need cards. Lets just list the elements. more like this: 
image
- we should add transfers here. If the wallet received or sent vault tokens.
- lets change "Activity" to "Your Activity"
- The withdraw dark red icon doesn't look great in dark mode. its pretty muddy. Lets talk to Al (UX) about color here.
- "Status" in the expanded section probably isn't necessary. What are the options other than "indexed"? 
- I think we should add the chain icon to the unexpanded chain name. maybe also the underlying logo to the vault name (but maybe not since the token logo is on the right).

rossgalloway (approved):
I think this is ready to go

PR #1206, opened 2026-05-04
feat: add codex wallet

w84april:
Add wallet impersonation for codex


w84april:
cherry picked into https://github.com/yearn/yearn.fi/pull/1264

PR #1209, opened 2026-05-05
add yvusd portfolio charts

rossgalloway:
## Description

Adds code to aggregate yvUSD charts to the portfolio page in the expanded rows and to the vault page in the charts section. The chart combines both locked and unlocked user positions and growth.

## Related Issue

adds the same functionality for yvUSD that was implemented for other vaults in #1190 

## Motivation and Context

May yvUSD charts equivalent to other vaults

## How Has This Been Tested?

To test:
- Run tests, build, and preview
- In a wallet with yvUSD balances, confirm that the yvUSD list row element on the portfolio page, when expanded, shows a combined balance and growth chart like the other rows.
- In a wallet with yvUSD balances, make sure the combined balance and growth chart shows up in the charts panel on the yvUSD vault page.

### DIFF
--- a/api/holdings/history.ts
+++ b/api/holdings/history.ts
@@ -35,6 +35,62 @@ function isValidAddress(address: string): boolean {
   return /^0x[a-fA-F0-9]{40}$/.test(address)
 }
 
+function parseVaultFilters({
+  vault,
+  chainId,
+  vaults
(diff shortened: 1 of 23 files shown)


w84april (approved):
lg!

PR #1237, opened 2026-05-19
feat: init redis

w84april:
## Description

Migrate from PostgreSQL to Redis


rossgalloway:
A review of the release branch (doesn't include this branch) flagged the following issue:

### Issue:
`api/lib/holdings/db/connection.ts` can fail legacy cache migration for mixed-case addresses.
During migration it lowercases `row.user_address`, then uses that lowercased value in a case-sensitive `WHERE user_address = $2`. Existing checksum/mixed-case rows won’t update, leaving empty `user_address_hash` values and causing the later unresolved-hashes check to fail.

### Potential solution:
**Use the original legacy address when updating migrated cache rows**

In `api/lib/holdings/db/connection.ts`, keep lowercasing only for the hash, but match the original stored address in the `WHERE` clause:

```ts
for (const row of legacyAddressesResult.rows) {
  const userAddressHash = toUserAddressHash(row.user_address)
  await db.query(
    `UPDATE holdings_totals
     SET user_address_hash = $1
     WHERE user_address = $2
       AND (user_address_hash IS NULL OR user_address_hash = '')`,
    [userAddressHash, row.user_address]
  )
}
```

Alternatively, use `LOWER(user_address) = $2`, but matching the exact original row value is the smaller, clearer migration fix.

Please check this and fix it in this branch if its an issue.

rossgalloway (approved):
LGTM

PR #1245, opened 2026-05-20
[codex] Harden Enso route execution

rossgalloway:
## Summary
- Add shared Enso route invariants for executable route payloads
- Require route tx targets to match the known Enso router for the request chain
- Require route tx sender and chain ID to match the request context
- Recheck the same invariants immediately before wallet transaction execution
- Validate successful Enso proxy responses before returning them from both API route implementations

## Security rationale
Enso route responses previously became executable wallet transactions after only basic shape checks. A tampered or malicious upstream response could choose tx.to, tx.data, tx.value, and tx.from. This change treats the upstream response as untrusted until it matches local request context and the local Enso router allowlist.

## Validation
- bun run lint:fix
- bun run tslint
- bunx vitest run src/components/pages/vaults/hooks/solvers/ensoRoute.test.ts src/components/pages/vaults/components/widget/deposit/useFetchMaxQuote.test.ts
- bun run build
- bun --print "await import('./api/enso/route.ts'); 'api route import ok'"

### DIFF
--- a/api/enso/route.ts
+++ b/api/enso/route.ts
@@ -1,4 +1,6 @@
 import type { VercelRequest, VercelResponse } from '@vercel/node'
+import { isAddress } from 'viem'
+import { normalizeEnsoRouteResponse } from '../../src/components/pages/vaults/hooks/solvers/ensoRoute'
 
 const ENSO_API_BASE = 'https://api.enso.finance'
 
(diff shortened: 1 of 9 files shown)


rossgalloway:
Closing this PR because it has been superseded by the grouped remediation flow.

The Enso route hardening is now covered by the grouped Enso proxy and transaction integrity remediation.

Replacement:
- https://github.com/yearn/yearn.fi/pull/1249 on `codex/fix-yearn-fi-group-enso-proxy-transaction-integrity`, head commit `e8a93e78`

PR #1257, opened 2026-05-23
feat: init vercel rate limit

w84april:
## Description

Vercel rate limit

Currently set to 50 rpm in vercel firewall settings as 10 rpm makes /activity page unusable. This is not a fix but shrinks the room for a bug, we should work on that


rossgalloway:
> This is not a fix but shrinks the room for a bug, we should work on that

what does this mean? Still good to review and merge?

PR #1265, opened 2026-06-02
chore: remove fixed rate KAT rewards after program end

rossgalloway:
## Description

This PR removes the `Base Rewards APR` row from the Katana APY breakdown modals as that program has now ended. The only rewards are now the rewards that flow through underlying strategies.

## Related Issue

https://github.com/yearn/katana-apr-service/pull/41

## Motivation and Context

Fix to keep site accurate

## How Has This Been Tested?

To Test:
- Open preview site.
- Filter for katana vaults. click on apy value to get popover modal. confirm no Base Rewards APR row shows.
- Navigate to Katana vault page. click on estimated or historical APY value in header. make sure Base Rewards APR row does not show.

image


w84april (approved):
lg

PR #1267, opened 2026-06-03
fix wallet connect connector dependency

rossgalloway:
## Description

This fix pins an older version of the qr dependency that the rainbowkit wallet connect qr code needs in the connection modal. This is a known issue with wevm's QR code library. Once it is fixed upsteam in that library we can remove this pinned dependency.

## Related Issue

https://github.com/wevm/cuer/issues/12
https://github.com/wevm/cuer/pull/11

## Motivation and Context

wallet connect connector is broken

## How Has This Been Tested?

tested locally. To test, open the preview site and try connecting a wallet via wallet connect. The QR code should load successfully.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "yearnfi",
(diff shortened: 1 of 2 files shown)


w84april (approved):
lg

PR #1282, opened 2026-06-17
fix: usdt approve

w84april:
## Description

USDT on Mainnet doesn't allow setting allowance if current allowance is not 0:
```
require(!((_value != 0) && (allowed[msg.sender][_spender] != 0)));
```

added a special message explaining that and cta

to test:
1. In any vault other than USDT enter amount that is bigger than your enso allowance
2. Message should appear
3. Click manage approval
4. Click revoke
5. Widget should refresh and it should be possible to deposit

image

### DIFF
--- a/src/components/pages/vaults/components/widget/deposit/ApprovalOverlay.tsx
+++ b/src/components/pages/vaults/components/widget/deposit/ApprovalOverlay.tsx
@@ -20,6 +20,7 @@ type TxState = 'idle' | 'confirming' | 'pending' | 'submitted' | 'success' | 'er
 interface ApprovalOverlayProps {
   isOpen: boolean
   onClose: () => void
+  onAllowanceUpdated?: (amount: bigint) => void
   tokenSymbol: string
   tokenAddress: `0x${string}`
   tokenDecimals: number
@@ -32,6 +33,7 @@ interface ApprovalOverlayProps {
 export const ApprovalOverlay: FC<ApprovalOverlayProps> = ({
   isOpen,
   onClose,
+  onAllowanceUpdated,
   tokenSymbol,
   tokenAddress,
   spenderAddress,
@@ -41,6 +43,7 @@ export const ApprovalOverlay: FC<ApprovalOverlayProps> = ({
 }) => {
   const [txState, setTxState] = useState<TxState>('idle')
   const [errorMessage, setErrorMessage] = useState('')
+  const [submittedAmount, setSubmittedAmount] = useState<bigint | undefined>()
 
   const { address: account, chain, connector } = useAccount()
   const currentChainId = useChainId()
@@ -82,6 +85,7 @@ export const ApprovalOverlay: FC<ApprovalOverlayProps> = ({
     if (!isOpen) {
       setTxState('idle')
       setErrorMessage('')
+      setSubmittedAmount(undefined)
       reset()
     }
   }, [isOpen, reset])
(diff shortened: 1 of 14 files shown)


murderteeth (dismissed):
works well, one small issue. here's what i did,

- open the usdc-1 vault page
- select usdt as the deposit token
- enter 1 usdt to deposit
- approve, but cancel deposit
- enter 2 usdt to deposit
- i see the "USDT approval needs reset" bubble
- click Manage approval
- Revoke
- ✅ enso router approval goes back to zero
- enter 2 usdt to deposit
- normal Approve x Deposit flow is enabled

issue: after the revoke is signed and confirmed, the deposit component still shows the "USDT approval needs reset" state for a couple seconds. a refresh happens. the deposit component reflects latest state again. so there's a state flicker there. not blocking, functionality is fixed. but feels janky

murderteeth (dismissed):
approved. previous reported state flicker is fixed, ty!

one more non-blocking note i'm noticing now:
- i enter 1 usdt to deposit
- i approve it but cancel the deposit
- i enter 2 usdt to deposit
- i get the usdt approval alert, but underneath i see the approve x deposit button still spinning

very small thing. i only point it out in case it reveals something else that might not be working as expected.

Image

murderteeth (changes requested):
### issue
when i have a larger deposit than allowance and i `Manage approval` i have options to `Revoke` and `Set Unlimited`. Revoke works but Set Unlimited fails to simulate and will fail if signed
Image

Image

### suggestions
- i click `Manage approval`, it opens a `Token Approval` dialog. the dialog and the link should have the same title. i like Manage approval. 
- don't mix sentence-case and title-case. for example `Manage approval` is sentence case and `Token Approval` is title case. Rule of thumb: use title case when it's a proper name of a person thing or place, use sentence case otherwise. So use title case for a vault name, use sentence case for dialog labels.
- i went through the usdt flow again. this time i imagined i was a user that didn't know usdt has this unusual allowance mechanism. two things were unclear to me on this screen. 1 why is usdt different? (why dont i need revoke with all erc20s).  2 what happens after i revoke? the Manage approval dialog tells you, but i already forgot by the time i read the revoke label. 
- what do you think of changing it from "Revoke" to "Reset". reset feels more user friendly to me.
Image

w84april:
@murderteeth 
1. Thx for pointing this. Don't think it's critical but probably worth adding some explanation UX wise:
image

```
        // To change the approve amount you first have to reduce the addresses`
        //  allowance to zero by calling `approve(_spender, 0)` if it is not
        //  already 0 to mitigate the race condition described here:
        //  https://github.com/ethereum/EIPs/issues/20#issuecomment-263524729
        require(!((_value != 0) && (allowed[msg.sender][_spender] != 0)));
```

2. Made copy to be `Manage approval` both in modal and button
3. Well it might be wordy but should be clear at least:
image

4. Reset might be friendly but I'd stick to Revoke as it seems to be accepted term e.g https://revoke.cash/

murderteeth (dismissed):
approved

regarding the usdt explainer text, i was thinking it would go in the modal actually where there's lots of available space to explain and the message gets anchored with the actual Revoke button. this way the initial alert message isn't as cramped too. 

Image

also 
> Revoke sets the Enso Router allowance to zero; then approve again.

reads better as 
> Revoke sets the Enso Router allowance to zero; then you can approve again.

murderteeth (approved):

murderteeth (approved):

PR #1290, opened 2026-06-26
feat: init calldata validation

w84april:
## Description

Validate calldata


rossgalloway:
@w84april what is this PR?

PR #1292, opened 2026-06-26
feat: add merkl key

w84april:
## Description

Add merkl key

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -42,6 +42,7 @@ VITE_CLUSTERS_API_KEY=
 # Enso reduces RPC load by fetching all balances via their API in a single call
 VITE_BALANCE_SOURCE=
 ENSO_API_KEY=
+MERKL_API_KEY=
 VITE_PLAUSIBLE_TRACK_LOCALHOST=
 
 # Comma-separated extra Vite dev/preview hosts. Keep personal hostnames in local .env only.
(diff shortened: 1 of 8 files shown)


matheus1lva (approved):

murderteeth (changes requested):
**Requesting changes — changes look good, but want to agree on approach before more gets built. If you disagree, push back is welcome.**

**1. Document your code changes in your commit messages, not just static docs.**
Static docs (README, etc.) are fine and welcome. But when the change is in *code*, I want the rationale in the commit message. The most important reason: the commit travels with the code forever via `git blame` — whoever touches this in six months sees *why* directly, with no archaeology through PRs and no dependence on a separate doc having been kept in sync. As-is, `feat: add merkl key` says nothing about the server proxy / secret-handling change underneath it.

You don't have to write these by hand — for example we often use this commit skill: https://github.com/yearn/webops-skills/blob/main/skills/create-commit/SKILL.md

**2. This endpoint needs rate limiting — via WAF, documented at `docs/firewall.md`.**
`/api/merkl/rewards` is public and now spends a metered `MERKL_API_KEY` on every call — before this PR there was no key to abuse, so without a limit anyone can drain the quota / run up the bill. I want the limit configured as a Vercel Firewall (WAF) rule only, and recorded in this repo at `docs/firewall.md` (route, limit, owner). No custom rate-limiting code unless it's absolutely necessary.

**3. Make a plan to remove the custom rate limiting from the holdings API.**
The holdings routes rate-limit with custom in-code logic (`api/lib/holdings/services/ratelimit.ts`, via the `@vercel/firewall` SDK). You flagged this yourself in #1257 — "this is not a fix… we should work on that." Same principle as above: please make a plan to drop the custom code and move holdings to WAF-only rate limiting. Let's use `docs/firewall.md` as the source of truth for the holdings rule and every other rate limit / firewall rule in this project.

**How do you want to proceed?**

murderteeth (approved):

PR #1302, opened 2026-07-06
Update 30d APY chart to use 30d interval by default

w84april:
## Description

Update 30d APY chart to use 30d interval by default
image


rossgalloway:
Can we change this to only affect the yvUSD vault and instead set yvUSD to 90-day as the default?

This solves the issue of the "big hump" in the data and still gives the longest possible dataset without it. And most charts are more informational over a full year so it seems like a shame to set it to 30 day for every vault when only one has an issue.

rossgalloway (approved):
LGTM

PR #1303, opened 2026-07-06
Fix Enso route slippage handling

w84april:
### Summary

This updates the vault deposit/withdraw Enso flow so the executable route is fetched with the user’s configured zap tolerance instead of starting from a `slippage=0` route.

Previously, the widget fetched an Enso route with `slippage=0`, used local USD prices to estimate total impact, then attempted a second fetch with the remaining tolerance. That first response still contained executable calldata, so in some paths the widget could end up signing a route where `amountOut === minAmountOut`. Any tiny state drift could then revert the transaction at Enso’s final slippage check.

Now the Enso route used by the widget is protected from the start with the configured slippage. The widget also reads Enso’s `priceImpact` field, applies `minAmountOut` consistently for Enso routes, and updates the warnings/details copy to talk about route price impact instead of swap-only slippage.

### Before

- Enso deposit/withdraw routes could be fetched first with `slippage=0`.
- The widget calculated total impact client-side from token prices, PPS, expected output, and min output.
- A second Enso request was supposed to add the remaining tolerance.
- Some routes could still expose or execute the zero-slippage quote.
- Price impact warnings were tied mostly to local USD valuation and swap detection.

### Now

- Enso routes are fetched with the configured zap tolerance directly.
- Enso `priceImpact` is parsed and used for route price impact warning/blocking when available.
- `minAmountOut` is used for Enso receive-at-least display across Enso routes, not only routes detected as swaps.
- Route warnings apply to Enso routes generally, not just `routeHasSwap`.
- Deposit/withdraw details now distinguish “receive” vs “receive at least” through an explicit `usesMinExpectedOut` flag.
- Obsolete double-fetch calibration code and related tests were removed.

### How to review

Start with the Enso route path:

- `src/components/pages/vaults/hooks/solvers/ensoRoute.ts`
- `src/components/pages/vaults/hooks/solvers/useSolverEnso.ts`
- `src/components/pages/vaults/hooks/actions/useEnsoDeposit.ts`
- `src/components/pages/vaults/hooks/actions/useEnsoWithdraw.ts`

Then review the widget behavior:

- `src/components/pages/vaults/components/widget/deposit/index.tsx`
- `src/components/pages/vaults/components/widget/withdraw/index.tsx`
- `src/components/pages/vaults/components/widget/deposit/DepositDetails.tsx`
- `src/components/pages/vaults/components/widget/withdraw/WithdrawDetails.tsx`

### Test plan

- [x] `bun run lint`
- [x] `bun run tslint`
- [x] Focused Enso/widget tests
- [ ] Manual: deposit with an Enso zap and confirm the executable route has `minAmountOut < amountOut` when slippage is non-zero.
- [ ] Manual: confirm high route impact blocks the action and settings still require acknowledgement above 1%.

### Risk / impact

This touches transaction construction for vault zaps, so the main risk is user-facing route gating.

This removes the unsafe zero-slippage executable route path, which should reduce random Enso runtime reverts caused by tiny output drift. However, the current implementation does not fully preserve the old PR #1158 “combined budget” semantics: Enso execution slippage is currently the full user tolerance, while Enso `priceImpact` is used separately for route impact warning/blocking. A follow-up can restore strict combined-budget behavior by fetching a quote-only estimate first, subtracting Enso `priceImpact`, and only exposing the final executable route.


w84april:
This solution is clean but doesn't include merged slippage and price impact like #1306 does, so should be closed after we merge 1306. (keeping it open just for tests)

w84april:
Closed in favour of #1306

PR #1306, opened 2026-07-07
Fix Enso protected quote slippage

w84april:
### Summary
Preserves the existing double-fetch Enso quote flow while fixing the zero-slippage execution risk.

Previously, the widget fetched a `slippage=0` Enso quote for calibration and then re-quoted with remaining slippage, but parts of that flow were gated by `routeHasSwap`. That meant non-swap Enso routes could skip the protected re-quote path and potentially expose zero-slippage calldata. This PR makes the bootstrap quote non-executable, applies min-output and route-impact checks to all Enso routes, and keeps the old combined tolerance model.

### How to review
Start with the deposit and withdraw widget quote state:
- `src/components/pages/vaults/components/widget/deposit/index.tsx`
- `src/components/pages/vaults/components/widget/withdraw/index.tsx`

Key behavior:
- first Enso quote still starts with `slippage=0`
- the widget waits for a protected quote before enabling execution
- Safe batches do not include Enso tx data while only the bootstrap quote is available
- remaining Enso slippage is still derived from the user tolerance minus estimated route impact
- `routeHasSwap` is now only used for swap-specific UI copy, not route-quality gating

Also review:
- Enso `priceImpact` parsing in `ensoRoute.ts`
- min-output display changes in `DepositDetails` / `WithdrawDetails`
- settings and warning copy updates

### Test plan
- [x] Automated: `bun run test -- src/components/pages/vaults/hooks/solvers/ensoRoute.test.ts src/components/pages/vaults/components/widget/deposit/DepositDetails.test.tsx src/components/pages/vaults/components/widget/withdraw/WithdrawDetails.test.tsx src/components/pages/vaults/components/widget/shared/PriceImpactWarning.test.tsx src/components/shared/utils/slippage.test.ts`

Manual smoke test:
- Deposit/withdraw through an Enso route and confirm the Network tab shows a `slippage=0` bootstrap request followed by a protected request with remaining slippage.
- Confirm the action button stays in fetching/disabled state until the protected quote is ready.
- Confirm high worst-case route impact blocks the transaction and shows the warning.

### Risk / impact
Touches zap execution readiness and displayed route-impact numbers for deposit/withdraw. The main risk is being stricter than before on Enso routes, especially routes that previously bypassed checks because they had no swap step. Rollback is scoped to the widget quote-state and Enso route parsing changes.

### DIFF
--- a/src/components/pages/vaults/components/widget/MobileDrawerSettingsButton.tsx
+++ b/src/components/pages/vaults/components/widget/MobileDrawerSettingsButton.tsx
@@ -72,7 +72,8 @@ export const MobileDrawerSettingsButton: FC = () => {
           <div className="space-y-1">
             <h4 className="font-medium text-text-primary">Transaction Settings</h4>
             <p className="text-xs text-text-secondary">
-              Applies site-wide across all vaults. Price impact covers both quote quality and execution buffer.
+              Applies site-wide across all vaults. Route impact consumes part of this tolerance; the remainder is used
+              as execution buffer.
             </p>
           </div>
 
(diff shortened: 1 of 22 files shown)


rossgalloway (approved):
LGTM

PR #1312, opened 2026-07-14
Fix failed transaction activity

rossgalloway:
## Summary

Failed Transactions were showing up as successful in the activity tab of the portfolio page. Upon deeper review, this was only for failed transactions in the local cache, which were not being tagged as "failed".

- require an explicitly successful receipt before transaction overlays complete
- treat reverted receipts as errors in transaction and approval flows
- revalidate cached local portfolio activity against on-chain receipts
- show reverted local activity with a `Failed` badge
- add regression coverage for reverted receipts and cached activity validation

## Root cause

The transaction overlays used the receipt query's `isSuccess` state as the transaction outcome. That state only means the receipt request completed; the receipt can still have a `reverted` status. Those transactions were persisted as successful local notifications, and the portfolio activity page trusted the cached status.

## Impact

Failed transactions are no longer recorded as completed. Existing cached success entries are checked against their receipts before display, and reverted entries remain available for auditability with a clear failed state. Indexed activity remains unchanged because reverted transactions do not emit the event logs used by the indexer.

## Validation

- `bun run lint:fix`
- `bun run tslint`
- focused Vitest suite: 61 tests passed
- `bun run build`


w84april (approved):
lg

PR #1315, opened 2026-07-16
Show weighted yvUSD portfolio APY

rossgalloway:
## Summary

Keep locked and unlocked yvUSD deposits combined in one portfolio row while showing the APY the user is actually earning.

- Weight locked and unlocked APYs by each position value
- Show a hover and tap breakdown with rates, weights, and USD values
- Use the personalized blended APY for portfolio metrics and sorting
- Add a dotted underline to communicate the APY tooltip affordance

## Root cause

The portfolio model already calculated a value-weighted yvUSD APY, but the list row rendered the synthetic unlocked vault APY instead. This caused locked-only and mixed positions to display a rate that did not reflect the user position.

## User impact

Users now see one yvUSD position with an APY matching their locked and unlocked allocation. The breakdown remains available without adding separate portfolio rows.

## Validation

- `bun run lint:fix`
- `bun run tslint`
- Focused Vitest coverage for APY weighting, tooltip content, and row rendering
- `bun run build`


w84april (approved):
looks great

PR #1318, opened 2026-07-16
Use 7-day APY for yBOLD estimates

rossgalloway:
## Summary

- show the 7-day historical net APY as the estimated APY for yBOLD and ysyBOLD
- apply the override to vault lists, detail pages, sorting, and portfolio calculations
- add tests covering both yBOLD product addresses and unchanged behavior for other vaults

## Why

As requested by Schlag and Johnny.
The forward estimate currently shown for yBOLD products does not reflect the desired user-facing yield. Both products should use their realized 7-day net APY wherever the interface presents an estimated or current APY.

## Validation

- `bun run tslint`
- `bun run lint:fix`
- `bunx vitest run src/components/shared/utils/vaultApy.test.ts src/components/pages/vaults/domain/normalizeVault.test.ts`
- `bun run build`


murderteeth (commented):
what about `based on underlying markets`  -> `based on 7 day historical performance`. "underlying markets" isn't accurate here

rossgalloway:
> what about `based on underlying markets` -> `based on 7 day historical performance`. "underlying markets" isn't accurate here

updated and pushed changes

murderteeth (approved):

PR #1319, opened 2026-07-17
Remove inactive yBOLD Pendle market

rossgalloway:
## What changed

- remove yBOLD from the fixed-term Pendle market mapping
- remove the stale Pendle icon and tooltip from the yBOLD vault list item and vault detail header
- preserve the active Pendle ARB rewards market mapping
- add regression coverage for both mappings

## Why

The yBOLD Pendle market is no longer active, but its vault address remained in the shared fixed-term market table. Both vault surfaces derived their Pendle affordances from that stale entry.

## Validation

- `bunx vitest run src/components/pages/vaults/constants/fixedTermMarkets.test.ts src/components/pages/vaults/components/table/apyDisplayConfig.test.tsx`
- `bun run tslint`
- `bun run lint:fix`
- `bun run build`


engn33r (approved):
Fixes the issue of the site linking to an expired Pendle market :+1:

w84april (approved):

PR #1338, opened 2026-08-04
Preserve Safe wallet connection on portfolio

rossgalloway:
### Summary
Keep the vault and portfolio routes under one shared app layout so the Wagmi and Safe provider tree remains mounted during client-side navigation.

This prevents a connected Safe iframe session from being reset when moving from `/vaults` or a vault detail page to `/portfolio`, while preserving the existing public URLs and route metadata.

### How to review
Start with `app/(wallet)/layout.tsx`, then confirm that the portfolio and vault route files moved beneath the shared route group without behavioral changes.

Smoke test in the Safe App embedding:
1. Connect a Safe on `/vaults` or a vault detail page.
2. Navigate to Portfolio using the site navigation.
3. Confirm the connected address remains visible and portfolio balances load.

### Test plan
- [x] Manual: production-mode `/vaults` to `/portfolio` navigation preserved the existing app/provider subtree with no browser errors
- [x] Automated: `bun run test` (171 files, 909 tests)

### Risk / impact
Low and limited to App Router layout ownership. Public paths, page components, metadata, wallet transaction code, and data-fetching behavior are unchanged. Reverting the single commit restores the previous per-route provider layouts.


w84april (approved):
tested lgtm

PR #1340, opened 2026-08-06
Codex/standalone deposit withdraw widget

rossgalloway:
## Description

fully standalone deposit withdraw widget. WIP.

might be horribly over-engineered


rossgalloway:
most of this is not in prod

PR #978, opened 2026-01-23
Use Enso API for balance fetch

murderteeth:
Integrates Enso Finance API for balance fetching and swap routing, reducing RPC load and improving security by proxying all Enso calls through the server.

### Balance Fetching
- `useEnsoBalances` - fetches balances via Enso API
- `useBalancesCombined` - hybrid approach: Enso for supported chains, multicall fallback for Fantom (250)
- `/api/enso/balances` - server proxy for balance API

### Swap Routing  
- `/api/enso/route` - server proxy for route API (removes client-side `VITE_ENSO_API_KEY`)
- Updated `useSolverEnso` and `useFetchMaxQuote` to use proxy

### Infrastructure
- Concurrent dev server (Vite + Bun API server via `bun dev`)
- Single `ENSO_API_KEY` env var for server-side auth

## Architecture

```
┌─────────────────────────────────────────────────────────┐
│                  useBalancesCombined                    │
├─────────────────────────────────────────────────────────┤
│  tokens                                                 │
│       ├──► Enso-supported chains ──► /api/enso/balances │
│       └──► Fantom (250) ──► multicall (RPC)             │
└─────────────────────────────────────────────────────────┘
```

## Configuration

```env
ENSO_API_KEY=           # Server-side only (required)
BALANCE_SOURCE=         # 'enso' (default) or 'multicall'
```

## Test plan

- [x] Connect wallet, verify balances load for all chains
- [ ] Verify Fantom balances load via multicall
- [ ] Test deposit/withdraw zaps still work
- [ ] Set `BALANCE_SOURCE=multicall`, verify fallback works (if testing in dev)

PR #1342, opened 2026-08-07
test: add regression test for portfolio double-counting (#1013)

### LINKED ISSUE #1013: Bug: Incorrect Portfolio total balance
### What happened 🆘

After merging enso balances PR (https://github.com/yearn/yearn.fi/pull/978) we have portfolio total balance calculated incorrectly. All the holdings are correct, this is calculation issue 

### To Reproduce 🔂

- Go to portfolio
- Connect wallet

BhariGowda:
Closes #1013

## What happened

After #978 merged, portfolio total balance was reported as calculated
incorrectly (#1013). A fix was attempted in #1016 (a targeted guard in
the old `apps/lib/contexts/useWallet.tsx`), but it was closed unmerged.

@rossgalloway asked in #1013: *"this is the staked asset balance
issue?"*  that question was never answered.

## What I found

Since #1013 was filed, the wallet/portfolio logic was rewritten
(`useWalletVaultTotals.ts`), and it now includes a `countedVaults` Set
that dedupes by vault identity  this correctly prevents the exact
double-count scenario described in #1013 (a vault counted once via its
raw share token, once via its staking token).

Confirmed rossgalloway's suspicion was correct, and the fix is already
in place  but **no test covers this exact scenario**. Only the
"hidden vault" exclusion paths are tested.

## What this PR does

Adds one regression test that encodes the original #1013 scenario: a
**visible** vault held via both its vault-share token and staking
token simultaneously should be counted once, not twice.

**Verified with a negative control**  temporarily removing the
`countedVaults` dedup guard makes this exact test fail with the
precise reported symptom (total `1000` instead of `500` for a single
$500 position). Restoring the guard passes it again.

Full suite: 911/911 passing locally (2 pre-existing unrelated failures
from a missing `@testing-library/dom` module, present before this
change too).

### DIFF
--- a/src/components/shared/contexts/useWalletVaultTotals.test.tsx
+++ b/src/components/shared/contexts/useWalletVaultTotals.test.tsx
@@ -8,6 +8,8 @@ const VISIBLE_VAULT_ADDRESS = '0x0000000000000000000000000000000000000001' as co
 const HIDDEN_VAULT_ADDRESS = '0x0000000000000000000000000000000000000002' as const
 const HIDDEN_STAKING_ADDRESS = '0x0000000000000000000000000000000000000003' as const
 const ASSET_ADDRESS = '0x0000000000000000000000000000000000000004' as const
+const STAKED_VISIBLE_VAULT_ADDRESS = '0x0000000000000000000000000000000000000005' as const
+const STAKED_VISIBLE_STAKING_ADDRESS = '0x0000000000000000000000000000000000000006' as const
 
 const { mockGetVaultHoldingsUsd, mockUseWalletTokens, mockUseYearn } = vi.hoisted(() => ({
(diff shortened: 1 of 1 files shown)


BhariGowda:
Thanks for checking in on this  makes sense given #1028 resolved the underlying issue back in February.

For the record, before opening this I'd verified the double-counting scenario from #1013 was still reachable on current `main` and confirmed the fix (or lack of one) with a negative control  removing the dedup guard reproduced the exact reported symptom, restoring it fixed it. Should have also checked recently-merged PRs for a prior fix on the same bug before starting, which would've surfaced #1028 sooner  noting that for next time.

If it'd be useful as a regression guard tied to #1028's fix, happy to reopen against that instead. Otherwise no action needed on my end  closing is the right call.

PR #1346, opened 2026-08-07
Remove unused dependencies and dead code

engn33r:
## Description

Removed old dependencies that weren't needed as well as dead code. Goal is to reduce the complexity of the codebase, which also reduces the attack surface for security problems.

## Related Issue

N/A

## Motivation and Context

Complexity reduction

## How Has This Been Tested?

Local dev server


rossgalloway:
## Summary

This PR removes six unused dependencies and approximately 6,500 lines of dead code across legacy vault components, hooks, utilities, ABIs, and server helpers. The retained application builds, tests, and renders successfully, including the vault list, vault detail, portfolio, fonts, charts, and filters.

## Dependencies

No dependencies were added. The PR removes `cross-fetch`, `ethers`, `graphql`, `vaul`, `@types/minimatch`, and `playwright`.

## Issues

None.

## Suggestions

- **src/components/shared/utils/constants.tsx:42** — Remove the stale V3 staking-zap comment now attached to `MERKLE_DISTRIBUTOR_ADDRESS`.
  - Change: Remove only the trailing `Address of the zap to deposit & stake for the v3 staking` comment.
  - Keep: Preserve the distributor address and the preceding Merkl distributor comment.
  - Done when: The retained constant is documented only as a Merkl distributor.

- **src/contexts/useDevFlags.tsx:20** — Consider completing the cleanup of the now-unreachable dev-flags API.
  - Change: Replace the remaining vault-detail consumer with the fixed `collapsible` mode and delete this context module, or retain and mount a real provider if runtime switching is still intended.
  - Keep: Preserve the current default collapsible header behavior.
  - Done when: The code no longer exposes unreachable display modes and no-op setters, or a mounted and tested provider makes those values reachable.

- **src/components/shared/components/MultiSelectDropdown.tsx:3** — Consider moving the retained option type out of the deleted component module.
  - Change: Move and rename `TMultiSelectOptionProps` to the vault-filter or shared-types area, remove the obsolete callback fields, and delete the type-only component file.
  - Keep: Preserve the label, value, selection state, and optional icon fields used by current consumers.
  - Done when: No live code imports a data model from the removed component path and the type contains only consumed fields.

## Verdict

APPROVE

---

## How This Was Reviewed
5 review lenses, each material finding independently verified by session. 0 candidate findings were refuted and dropped.

rossgalloway:
@w84april this looks pretty good to me, but if you could give it a review, that would be great.

matheus1lva (approved):

rossgalloway:
These changers were merged in #1363

PR #1348, opened 2026-08-07
Fix Enso cross-chain calibration

rossgalloway:
### Summary
Some Enso cross-chain deposits and withdrawals failed before the user could submit a transaction. This PR fixes the first quote that yearn.fi sends to Enso.

For a cross-chain route, yearn.fi now uses 0.01% slippage for the first quote. This first quote is used only to check the route. The app never lets the user execute it. The app then asks for a second quote using the user's slippage setting. Only the second quote can be executed.

Slippage is the maximum price change that the user is willing to accept.

### What was wrong
Users saw this error in both production and local builds:

> All cross-chain bridge routes failed. No valid route found.

This message made the problem look like a bridge outage. It was not a bridge outage.

Yearn.fi sends two quotes for an Enso transaction:

1. The first quote checks the route and measures its price impact.
2. The second quote uses the user's slippage setting and is the quote that can be executed.

Before this PR, the first quote used 0% slippage. Enso could not build some complex cross-chain routes with 0% slippage. Enso returned an error, so yearn.fi never asked for the second quote.

This problem was in the shared deposit and withdrawal code on `main`. It was not caused by the new swap page, and it was not limited to one chain.

We reproduced the problem with routes to both Base and Katana:

| Route | First quote setting | Result |
| --- | ---: | --- |
| Ethereum WETH to Katana Yearn vault | 0% | Failed with HTTP 400 |
| Ethereum WETH to Katana Yearn vault | 0.01% | Worked |
| Ethereum USDC to Base Yearn vault | 0% | Failed with HTTP 400 |
| Ethereum USDC to Base Yearn vault | 0.01% | Worked |

Some simple cross-chain token transfers worked with 0%. This is why the problem looked random at first.

### What this PR changes
For cross-chain routes:

1. The first quote now uses 0.01% slippage instead of 0%.
2. Yearn.fi uses the first quote to measure the route.
3. Yearn.fi calculates the allowed slippage from the user's setting.
4. Yearn.fi asks Enso for a second quote.
5. Only the second quote can be executed.

This PR does not increase the user's final slippage setting. The 0.01% value is only used for the first, non-executable quote.

If there is no safe slippage value left for the second quote, the app blocks the transaction and shows an error. Same-chain quotes continue to work as before.

### How to review
Start with `src/components/pages/vaults/hooks/solvers/useSolverEnso.ts`. Check that only the first cross-chain quote gets the 0.01% minimum.

Then review `useProtectedEnsoQuoteState.ts`. Check these safety rules:

- The first quote can never be executed.
- A loading or blocked quote can never be executed.
- Only the second quote can provide a transaction.
- The second quote never uses more slippage than the user allowed.

The deposit and withdrawal files only show the new error messages in the existing UI.

### Test plan
- [x] Automated: protected quote state, route request, and slippage tests (15 tests)
- [x] Automated: touched-file Biome check
- [x] Manual: Ethereum USDC to Base Yearn vault returned complete quotes at 1 bps and 100 bps
- [x] Manual: Ethereum WETH to Katana Yearn vault returned complete quotes at 1 bps and 100 bps
- [x] Manual: submit transaction

Note: `bun run lint:fix` does not scan this Codex worktree because its full path contains `.codex`. All seven changed files were checked directly with Biome.

### Risk / impact
This changes how yearn.fi asks Enso for cross-chain deposit and withdrawal quotes. It does not change contracts or move funds by itself.

The most important safety rule is that the first quote must never be executable. Tests cover this rule. The final quote still follows the user's slippage setting and the existing 5% safety limit.

Rollback is a single commit revert.

### DIFF
--- a/src/components/pages/vaults/components/widget/deposit/depositStepSuccess.ts
+++ b/src/components/pages/vaults/components/widget/deposit/depositStepSuccess.ts
@@ -1,0 +1,26 @@
+type HandleDepositStepSuccessParams = {
+  label: string
+  approvalFlowKey: string
+  refetchAllowance?: () => Promise<unknown>
+  completeApprovalFlow: (approvalFlowKey: string) => void
+}
+
+type AllowanceRefetchResult = {
+  error?: unknown
+}
+
+export async function handleDepositStepSuccess({
+  label,
+  approvalFlowKey,
+  refetchAllowance,
+  completeApprovalFlow
+}: HandleDepositStepSuccessParams): Promise<void> {
+  if (label !== 'Approve' && label !== 'Sign Permit') return
+
+  if (label === 'Approve') {
+    const result = (await refetchAllowance?.()) as AllowanceRefetchResult | undefined
+    if (result?.error) throw result.error
+  }
+
+  completeApprovalFlow(approvalFlowKey)
+}
--- a/src/components/pages/vaults/components/widget/deposit/depositStepSuccess.test.ts
+++ b/src/components/pages/vaults/components/widget/deposit/depositStepSuccess.test.ts
@@ -1,0 +1,71 @@
+import { handleDepositStepSuccess } from '@pages/vaults/components/widget/deposit/depositStepSuccess'
+import { describe, expect, it, vi } from 'vitest'
+
+describe('handleDepositStepSuccess', () => {
+  it('refreshes the confirmed allowance before completing an approval flow', async () => {
+    const events: string[] = []
+    const refetchAllowance = vi.fn(async () => {
+      events.push('allowance-refreshed')
+    })
+    const completeApprovalFlow = vi.fn(() => {
+      events.push('approval-completed')
+    })
+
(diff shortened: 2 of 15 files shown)


src/components/pages/vaults/components/widget/deposit/depositStepSuccess.ts
```diff
@@ -0,0 +1,26 @@
+type HandleDepositStepSuccessParams = {
+  label: string
```
matheus1lva, line 2: nit: 
```suggestion
  label: "Approve" | "Sign Permit" | string
```

matheus1lva (approved):

w84april (approved):
lgtm, tested works fine. thing to keep in mind - there could be an edge case when route query #1 and #2 return different price impact, so if #2 price impact if much larger, then slippage may be overflown. I have never seen 2 different price impacts tbh, so its ok imo

rossgalloway:
These changes were made in #1351

PR #1350, opened 2026-08-11
Feat/init monorepo

w84april:
### Summary

Convert the repository into a Bun monorepo and extract the production deposit/withdraw implementation into a private `@yearn/vault-widget` workspace package.

Yearn remains the primary consumer under `apps/yearn`, while the new `apps/ybold` application validates reuse with its own providers, services, styling, and deployment. The package is linked through `workspace:*` and is not published to npm.

This also introduces a framework-neutral transaction planner and injectable executor. Ready, same-chain, single-request EOA deposits and withdrawals use the new execution path when configured; approvals, Safe, Enso, cross-chain, chain-switch, and dynamic unstake flows retain the established production path.

### How to review

The commits are intentionally ordered by concern:

1. `e3f7844d` moves the existing Yearn application into `apps/yearn`. Most of this commit is mechanical renaming; focus on root workspace configuration, scripts, environment loading, and Vercel paths.
2. `2a8ea41a` extracts the production widget into `packages/vault-widget`. Start with:
   - `packages/vault-widget/package.json`
   - `packages/vault-widget/src/index.tsx`
   - `packages/vault-widget/src/runtime.tsx`
   - `apps/yearn/src/components/shared/contexts/YearnVaultWidgetRuntimeProvider.tsx`
3. `1ebf5318` adds yBOLD as a second consumer under `apps/ybold`.
4. Review the headless transaction path:
   - `packages/vault-widget/src/headless/buildTransactionPlan.ts`
   - `packages/vault-widget/src/headless/executeTransactionPlan.ts`
   - `packages/vault-widget/src/adapters/wagmiExecutionAdapter.ts`
   - the planned transaction controller and Deposit/Withdraw integrations
5. Review `.github/workflows/lint.yml` and `scripts/check-vault-widget-boundary.ts` for workspace and dependency-boundary enforcement.

Please pay particular attention to:

- approval/reset ordering and allowance handling;
- canonical-chain to execution-chain mapping;
- transaction receipt and hash validation;
- single-chain Safe enforcement and atomic batching;
- which flows enter the new executor versus the legacy fallback;
- reviewed yBOLD addresses and preset policy;
- preservation of Yearn notifications, callbacks, refreshes, and success UI.

### Scope boundaries

- The package is private and reused only through this repository’s workspace.
- Nothing is published to npm or another registry.
- The package currently exports TypeScript source and is transpiled by each Next application.
- Migration, rewards, and other Yearn-specific product flows remain owned by the Yearn app.
- This does not move every transaction family onto the new headless executor.
- No contracts, databases, or persistent data formats are migrated.

### Risk / impact

This PR touches transaction execution and should be treated as funds-sensitive.

The new executor is deliberately limited to already-prepared, same-chain EOA requests. More complex flows continue through the production state machine until their confirmation and resume behavior can be represented safely by the public transaction-plan contract. If a host does not provide a complete execution adapter, the widget uses the legacy path.

The repository layout change also affects deployment:

- The existing Yearn Vercel project must use `apps/yearn` as its Root Directory.
- yBOLD requires a separate Vercel project rooted at `apps/ybold`.
- Both projects must be allowed to access files outside their Root Directory so they can use the root lockfile and `packages/vault-widget`.


rossgalloway (commented):
## Summary

This PR converts the repository into a Bun monorepo, extracts the deposit and withdrawal widget into a private workspace package, adds yBOLD as a second consumer, and introduces a framework-neutral transaction planner and executor.

The package boundary, typechecks, local tests, production builds, and browser smoke tests passed. The current GitHub test job remains red under Node 20 because three tests call `Promise.withResolvers`; all 765 Yearn tests pass locally under Node 22.

## Dependencies

- **@next/env@16.2.11**: APPROVED — it preserves Next.js environment-loading behavior, matches the existing Next version, is small, maintained, MIT-licensed, and introduces no additional package version.
- **@wagmi/core@2.22.1**: APPROVED — the execution adapter directly uses its transaction APIs; it is already installed through Wagmi, supports ESM tree-shaking, is maintained, MIT-licensed, and introduces no duplicate version.

Bundlephobia returned HTTP 429, so exact minified and compressed sizes could not be recorded.

## Issues

- [ ] **packages/vault-widget/src/headless/buildTransactionPlan.ts:146** — Safe plans do not reject a target chain that differs from the connected Safe chain until simulation or wallet interaction.
  - Change: Before building or simulating the Safe proposal, require a known connected canonical chain that equals the proposal chain.
  - Keep: Mixed-chain Safe plans must remain rejected, and reset, approval, and execution calls must retain their current order in one atomic proposal.
  - Done when: A test rejects a chain-1 Safe intent while connected to chain 10 before `proposeSafeBatch`, while the matching-chain case produces the existing ordered batch.

- [ ] **packages/vault-widget/src/headless/executeTransactionPlan.ts:126** — The public executor accepts a successful receipt whose `transactionHash` does not match the submitted hash.
  - Change: Validate submitted hashes and require the EOA and Safe receipts to contain valid, matching transaction hashes before advancing or refreshing.
  - Keep: Reverted receipts and accumulated submissions must remain available in `VaultWidgetPlanExecutionError`, and receipt failures must continue to skip refresh.
  - Done when: Tests reject malformed or mismatched hashes before refresh or success, while matching receipts retain the existing state sequence.

- [ ] **packages/vault-widget/src/headless/buildTransactionPlan.ts:143** — A zero-call intent with no required approval becomes a refresh-only plan that reports success with no submissions.
  - Change: Reject an intent whose `calls` array is empty before constructing the plan, or enforce and validate a non-empty call contract.
  - Keep: Approval ordering, chain switching, Safe batching, execution ordering, and final refresh must remain unchanged for valid intents.
  - Done when: A deterministic test rejects an empty-call deposit or withdrawal intent, while existing non-empty plans remain unchanged.

- [ ] **apps/yearn/src/components/pages/vaults/hooks/actions/useDirectDeposit.ts:20** — The extraction leaves approximately 1,453 lines of unused app-local transaction hooks and solvers alongside the package implementation used in production.
  - Change: Export the required staking conversion helpers intentionally from the package, redirect all three live app imports, and then delete only the unused duplicated action and solver stack.
  - Keep: Keep the app-local `useTokenAllowance` used by `WidgetMigrate`, plus the app-owned migration and managed yvUSD cooldown flows.
  - Done when: Production deposit and withdrawal imports resolve to the package, no unused duplicate stack remains, and existing widget and Yearn tests still pass.

- [ ] **packages/vault-widget/src/components/widget/shared/TransactionOverlay.tsx:570** — No rendered component test exercises the new planned-execution branch or proves that an undefined plan reaches the legacy executor.
  - Change: Add rendered tests for one eligible planned transaction, pre-submission retry, and one representative undefined-plan legacy fallback.
  - Keep: Existing eligibility tests should continue proving that approvals and Safe inputs do not create plans; all legacy-only transaction families must remain on the legacy path.
  - Done when: Tests fail if the overlay starts the planned executor twice, loses notification/refresh/success callbacks, makes a submitted transaction retriable, or sends the representative fallback through the new executor.

## Suggestions

- Coverage not reached: verify both Vercel projects with `apps/yearn` and `apps/ybold` roots, outside-root workspace access, and representative deep links.
- Coverage not reached: add yBOLD host tests for APY fallback, spot-price proxy responses, mainnet wallet handling, and rendered deposit/withdraw states.
- Coverage not reached: run wallet-backed Tenderly smoke tests for eligible EOA deposit/withdraw and representative approval, Safe, and Enso legacy paths.
- Coverage not reached: smoke-test relocated Yearn redirects, API routes, assets, security headers, and root-loaded environment configuration.
- Coverage not reached: exercise the safe modes of the path-sensitive operational scripts changed for the monorepo layout.

## Verdict

COMMENT

---

## How This Was Reviewed

5 review lenses, each material finding independently verified by `session` using fresh Codex contexts. This was same-model independent-context verification, not cross-model verification. 0 candidate findings were refuted and dropped.

murderteeth (changes requested):
## Summary

Converts the repo into a Bun monorepo (`apps/yearn`, `apps/ybold`, `packages/vault-widget`), extracts the deposit/withdraw widget into a private workspace package, and adds a headless transaction planner/executor for ready same-chain EOA flows. Type checks and tests pass in all three workspaces (765 + 237 + 1), the boundary check passes, and both apps render and interact correctly in a browser smoke test.

## Dependencies

- **@next/env@16.3.1**: APPROVED — build-time-only Vercel package, already transitive via `next`, version in lockstep with `next` 16.3.1, no advisories.
- **@wagmi/core@2.22.1**: APPROVED — exact version already resolved via `wagmi` 2.18.2 (single deduped lockfile entry); declaring it directly fixes phantom imports in the new adapter. No advisories.

## Issues

- [x] **yBOLD can be clickjacked (high)** — `apps/ybold/next.config.ts:3` — any website can frame the new dapp and trick users into clicking its deposit/withdraw controls. The config sets no headers at all, while `apps/yearn/next.config.ts` applies a full `securityHeaders` set to `/:path*` in this same PR.
  - Change: add a `headers()` mirroring `apps/yearn/next.config.ts` with `frame-ancestors 'self'` (plus `https://app.safe.global` only if Safe embedding is intended); set `poweredByHeader: false`.
  - Keep: `transpilePackages` and the safe-apps-sdk turbopack alias unchanged.
  - Done when: deployed yBOLD responses carry CSP `frame-ancestors`, HSTS, nosniff, and Referrer-Policy, and omit X-Powered-By.

- [x] **Failed transactions look unverifiable (high)** — `packages/vault-widget/src/adapters/wagmiExecutionAdapter.ts:136` — when a transaction reverts on-chain in the new execution path, the user is told it "could not be verified" with no retry, and the pending notification never turns into a failure.
  - Change: in `waitForReceipt`, return the reverted receipt instead of throwing (keep hash/shape validation) so the executor's `requireSuccessfulReceipt` classifies it; update `wagmiExecutionAdapter.test.ts:160-163`, which codifies the throw.
  - Keep: malformed hashes/receipts still rejected; successful receipts unchanged.
  - Done when: an on-chain revert shows "Transaction failed" + Try Again and the pending notification updates to error.

- [ ] **"Speed up" in the wallet breaks a successful deposit (high)** — `packages/vault-widget/src/adapters/wagmiExecutionAdapter.ts:140` — if the user speeds up a pending transaction, it succeeds on-chain but the widget reports an error, never refreshes balances or shows success, and the notification stays pending.
  - Change: replace raw receipt-hash equality with viem's `onReplaced` information — accept a successful `repriced` replacement and track its new hash; treat `cancelled` as failure.
  - Keep: genuinely mismatched or malformed receipts still rejected.
  - Done when: a sped-up transaction completes the plan (refresh + success UI + notification success), a cancelled one reports failure, and a test covers the repriced case.

- [x] **Money-math helpers now maintained twice (medium)** — `apps/yearn/src/components/pages/vaults/hooks/actions/stakingAdapter.ts:1` — five funds-sensitive helpers (`stakingAdapter`, `useTokenAllowance`, `utils/yBold`, `utils/ensoRouters`, `utils/nativeTokens`) live as identical copies in both app and package, each with live consumers, so a fix to one will silently miss the other.
  - Change: replace the app-side bodies with re-exports from the package's `./internal/hooks/*` and `./internal/utils/*` subpaths.
  - Keep: `check-vault-widget-boundary` keeps passing — only the app → package direction is added.
  - Done when: single implementations, app consumers compile, `ensoRouters.test.ts` passes against the shared copy.

- [x] **Balance math duplicated across app and widget (medium)** — `apps/yearn/src/components/pages/vaults/hooks/useVaultUserData.ts:1` — the wallet panel and vault header use a ~220-line copy of the widget's user-data hook, so a future fix to deposited-value or staking math will reach one surface and not the other.
  - Change: consolidate on the package hook (the runtime provider is already mounted globally), or if the app copy must stay for tenderly virtual-net support, cross-reference both files and name the owner of the shared math.
  - Keep: WalletPanel/VaultDetailsHeader behavior including tenderly chain resolution; `VaultDetailsHeader.test.tsx` passes.

- [x] **Analytics chain breakdown will go dark (medium)** — `packages/vault-widget/src/components/widget/deposit/index.tsx:957`, `withdraw/index.tsx:966` — the deposit/withdraw events renamed their `chainID` property to `chainId`. Event names are unchanged, so existing Plausible breakdowns keyed on `chainID` will silently report zero from deploy onward and per-chain history splits across two names.
  - Change: keep emitting `chainID`, or confirm the rename is intentional and migrate the dashboards and any alerting in the same deploy.
  - Keep: event names `deposit`/`withdraw` and the remaining property keys stay as they are on main.
  - Done when: per-chain deposit/withdraw breakdowns show no gap at deploy time.

- [x] **~1,950 lines of dead code left behind (low)** — `apps/yearn/src/components/pages/vaults/hooks/actions/useDirectDeposit.ts:1` — sixteen transaction hooks left in apps/yearn have no consumers anymore; the next person to touch them will be editing code that never runs.
  - Change: delete the 16 files (`useDirect*`/`useEnso*`/`useYBold*`/`useYvUsd*` actions, `useSolverEnso`, `ensoRoute` + test, `useEnsoOrder`, `useInput`, `useDebouncedInput`).
  - Keep: `stakingAdapter.ts` stays — it still has live consumers (finding above).
  - Done when: files gone, `tslint` and the apps/yearn suite pass, grep matches only package imports.

## Verdict

REQUEST_CHANGES

---

## How This Was Reviewed
5 review lenses, each finding independently verified by codex. 3 candidate findings were refuted and dropped.

w84april:
@murderteeth thx for the review! fixed all mentioned except the one with "Speed up". I spent some time on it but didn't work out and i think this pretty rare case and not a big deal, so could go

murderteeth (changes requested):
## Summary

Monorepo split (`apps/yearn`, `apps/ybold`, `packages/vault-widget`) plus a headless transaction planner/executor. Both apps build and render correctly, all 1,005 tests pass, the boundary check holds, and the yBOLD addresses verify on-chain (`yBOLD.asset() == BOLD`, `ysyBOLD.asset() == yBOLD`). The yBOLD security-header fix is confirmed live in served responses.

Six of the items below are regressions — behavior that works on `main` today and stops working after this merge. Two were previously reported as fixed and are not.

## Issues

- [x] **Reverted transactions can no longer be verified (high) — regression from `main`** — `packages/vault-widget/src/adapters/wagmiExecutionAdapter.ts:132` — on `main`, `transactionOverlay.helpers.ts:12` maps `status === 'reverted'` to `error` and reverts surface correctly. The fix in `f3861d6b` doesn't restore that: `@wagmi/core`'s `waitForTransactionReceipt` throws on a reverted receipt itself, so the removed line was already unreachable and the adapter never returns a reverted receipt. The user gets "Transaction submitted / confirmation could not be verified" with **Close** and no retry, and the notification stays pending forever. The new test passes only because it stubs `waitForReceipt` with a value the real adapter cannot produce.
  - Change: read the receipt via `getPublicClient(config, {chainId}).waitForTransactionReceipt(...)` instead of the `@wagmi/core` action, so a reverted receipt reaches `requireSuccessfulReceipt`; drive the test through the real adapter with a mocked transport.
  - Keep: malformed hashes and receipts still rejected.
  - Done when: an on-chain revert shows "Transaction failed" + Try Again, the notification ends in error, and the test fails if the adapter is swapped back.

- [ ] **Wallet "Speed up" breaks a successful deposit (high) — regression from `main`** — `packages/vault-widget/src/adapters/wagmiExecutionAdapter.ts:139` — `main` never compared hashes, so viem's replacement receipt flowed through and sped-up transactions succeed today. This adapter adds a raw hash-equality check, so the transaction succeeds on-chain while the widget reports an unverifiable error, never refreshes balances, never shows success, and leaves the notification pending. The fix existed in `1c28d1f8` and was reverted in `91f913b6`.
  - Change: restore `1c28d1f8` — pass `onReplaced`, accept the mined receipt when `replacedTransaction.hash` matches the submitted hash, track the new hash, treat `cancelled` as failure.
  - Keep: receipts without valid replacement evidence still rejected.
  - Done when: a repriced transaction completes with refresh + success UI + notification success, and a test covers it.

- [x] **yBOLD's Next pin adds a second, vulnerable dependency tree (high) — new in this PR** — `apps/ybold/package.json:19` — pinning `next` to `16.2.11` while Yearn is on `16.3.0` creates a separate tree in `bun.lock` (`@yearn/ybold/next/postcss`, `@yearn/ybold/next/sharp`) carrying **postcss 8.4.31** and **sharp 0.34.5**. `main` already carried a transitive postcss 8.5.15, so postcss exposure is worsened rather than newly created — 8.4.31 falls below all three thresholds (GHSA-6g55-p6wh-862q high ≤8.5.11; GHSA-r28c-9q8g-f849 high ≤8.5.17; GHSA-fxqj-rqcc-2cmp medium ≤8.5.22). **sharp 0.34.5 is new exposure**: `main` resolved only 0.35.3, and GHSA-f88m-g3jw-g9cj (high) affects <0.35.0. Both are build/image-pipeline packages, not reachable through the deployed request path.
  - Change: pin `apps/ybold` to `next` `16.3.0` and regenerate `bun.lock`.
  - Done when: `bun.lock` contains no `postcss@8.4.31` or `sharp@0.34.5` entry.

- [x] **Base loses its extra confirmation (medium) — regression from `main`** — `packages/vault-widget/src/adapters/wagmiExecutionAdapter.ts:131` — `main` waits 2 confirmations on chain 8453 ("Fast chains like BASE need extra confirmations"); the planned path takes wagmi's default of 1, so Base deposits show success and refresh a confirmation early. The legacy path still waits 2, so the two paths now disagree.
  - Change: plumb a per-chain confirmations value into the adapter and apply the same `8453 => 2` rule.

- [x] **Gas buffer dropped (medium) — regression from `main`** — `packages/vault-widget/src/adapters/wagmiExecutionAdapter.ts:121` — `main` sends `gas: (gasEstimate * 110n) / 100n`; the adapter omits `gas` entirely and leaves the limit to the wallet, which may add no margin on state-sensitive calls.
  - Change: estimate with the public client already in hand at line 115, apply the same 10% margin, pass it to `sendTransaction`.

- [x] **Analytics chain breakdown goes dark (medium) — regression from `main`** — `packages/vault-widget/src/components/widget/deposit/index.tsx:957`, `withdraw/index.tsx:966` — reported fixed but unchanged: the Plausible property is still `chainId` where `main` sent `chainID`, and every other Yearn event still uses `chainID`. Event names are unchanged, so existing per-chain breakdowns silently report zero from deploy.
  - Change: restore `chainID`, or emit both for one release and migrate the dashboards first.

- [x] **Merged dependency upgrades are rolled back (medium) — regression from `main`** — `apps/yearn/package.json:32` — `origin/main` is the merge-base, so merging immediately downgrades react/react-dom 19.2.8→19.2.7, viem 2.55.11→2.55.8, `@upstash/redis` 1.38.2→1.38.0, `@tanstack/react-virtual` 3.14.9→3.14.8, `@types/react` 19.2.18→19.2.17, playwright 1.62.1→1.61.1, and postcss 8.5.26→8.5.19 — the last landing back inside GHSA-fxqj-rqcc-2cmp (medium, affected ≤8.5.22, patched 8.5.23). Confirmed in `bun.lock`.
  - Change: carry `main`'s pins across into `apps/yearn`, `apps/ybold`, and the widget's peer/devDeps; regenerate the lockfile.
  - Keep: exact pins only, so the npm-package-json-lint rules keep passing.

- [x] **yBOLD price proxy drops protections the Yearn route has (low) — new in this PR** — `apps/ybold/app/api/prices/spot/route.ts:19` — `coins` is forwarded unvalidated, and the 2-minute CDN cache applies to error responses. The equivalent Yearn route validates via `parseCoins` (400 on bad input) and sets `Cache-Control: no-store` when `!upstream.ok`, so one transient yearn.fi failure is served for minutes here.
  - Change: mirror `parseCoins` and the `no-store`-on-error behavior from `apps/yearn/src/server/prices/spot.ts`.

- [x] **Two cleanup items marked fixed did not land (low) — new in this PR** — `apps/yearn/src/components/pages/vaults/hooks/actions/useDirectDeposit.ts:1` — roughly 1,900 lines of unreferenced transaction hooks and solvers remain in `apps/yearn`, and eight live helper modules (`stakingAdapter`, `useTokenAllowance`, `useVaultUserData`, `yBold`, `nativeTokens`, `slippage`, `ensoRouters`, `approve`) still exist as parallel copies in app and package.
  - Change: delete the unreferenced files; re-export the shared eight from `@yearn/vault-widget` so each has one implementation.

## Verdict

REQUEST_CHANGES

---

## How This Was Reviewed
5 review lenses, each finding independently verified by codex. 2 candidate findings were refuted and dropped.

w84april:
@murderteeth fixed all the above except speed up (definitely not high prio + not a regression).
ran clanker on my side and it only pointed this + some low priority issues, so should be good imo

murderteeth (changes requested):
## Summary

Seven fix commits since 85915ecb. Eight of the nine issues from the last review are verified fixed — checked against the code, not the commit messages. All 1,007 tests pass, both apps build, and both render with live data.

**Every issue below is new**, introduced by one of these seven commits. Nothing carried over from the previous round.

## Issues

- [ ] **[HIGH] Transactions slower than 3 minutes now dead-end**
-- Swapping `@wagmi/core`'s action for `publicClient.waitForTransactionReceipt` silently inherited viem's default timeout. `@wagmi/core@2.22.1` destructures `timeout = 0` and forwards it, so viem builds no timer and waits indefinitely; viem 2.55.11 defaults `timeout = 180_000` and rejects with `WaitForTransactionReceiptTimeoutError`. The rejection lands after the hash is recorded but before a receipt exists, so the controller classifies it `submitted-unconfirmed`: "Transaction submitted / confirmation could not be verified", **Close**, no retry, notification stuck pending — for a transaction that confirms fine a minute later. Mainnet congestion reaches 3 minutes routinely. The adapter currently hands viem `{"confirmations":1,"hash":"0x…"}` and nothing else.
-- Introduced in `698555e5`; regression from `main`, where the legacy path also waited indefinitely. Four review lenses flagged it, three independent verifiers confirmed it.
-- Change: pass an explicit `timeout: 0` to `publicClient.waitForTransactionReceipt`, or a deliberately chosen value validated the way `safeTimeoutMs` already is; add a test asserting the forwarded timeout.
-- Keep: reverted receipts must still resolve rather than throw, so `requireSuccessfulReceipt` stays the single place a revert is classified.
-- Done when: a transaction confirming after 180s still resolves to its receipt, and a test pins the timeout the adapter sends.

- [ ] **[LOW] Genuine Safe failures show the wrong message**
-- The new Safe/generic message split never takes effect. The branch calls `resetTxState()`, which clears `txHash`; `txHash` was added to the same effect's dependency array in the same change, so the effect immediately re-runs. On re-entry `resolvePendingSafeOverlayState` returns `'error'` unchanged (`transactionOverlay.helpers.ts:100`), the branch runs again with `txHash` now undefined, `isConfirmedSafeTransactionFailure` short-circuits false, and the Safe message is overwritten with "Transaction failed. Please try again."
-- Introduced in `698555e5`; regression from `main`, which showed the Safe guidance for these failures. The change inverts the bug rather than fixing it.
-- Change: latch `isConfirmedSafeFailure` before `resetTxState()`, and early-return when `overlayState` is already `'error'` so the branch runs only on the transition into it.
-- Keep: the fix's actual goal — a non-Safe failure must still show the generic message, not the Safe queue copy.
-- Done when: a Safe transaction whose `safeTxStatus` is `failed` or `cancelled` still reads "Transaction failed in Safe" after the effect settles, and a test covers the re-entry.

- [ ] **[LOW] Enso route parsing lost all test coverage**
-- The cleanup deleted `apps/yearn/.../solvers/ensoRoute.test.ts` (159 lines) along with the app-local copy, but the package implementation it covered is still live: `useSolverEnso.ts` calls `routeHasSwapStep` at lines 109 and 194 and imports `normalizeEnsoRouteResponse` at line 17. `packages/vault-widget/src/hooks/solvers/` now contains no test file at all.
-- Introduced in `72beddb5`; regression from `main`, where the suite existed and ran.
-- Change: restore the suite as `packages/vault-widget/src/hooks/solvers/ensoRoute.test.ts`, importing from `./ensoRoute`.
-- Keep: the implementation is unchanged, so the restored tests need no edits.
-- Done when: `bun run test:widget` executes the Enso route assertions again.

- [ ] **[LOW] Dead normalization layer left behind**
-- Now that `useVaultUserData` re-exports the package hook, `normalizeVaultUserDataForWidget` receives tokens that already carry `chainId`, so the `'chainId' in token` branch always wins and its `fallbackChainId` argument is unreachable. It rebuilds objects to no effect at `advanced.tsx:31` and `:49`.
-- Introduced in `72beddb5`; new in this PR, not present on `main`.
-- Change: pass `vaultUserData` straight through and delete `normalizeVaultUserDataForWidget`.
-- Keep: `normalizeTokenForWidget` stays — `advanced.tsx:33` still feeds it a `TToken` carrying `chainID`.
-- Done when: no call site passes a `fallbackChainId` that cannot be reached, and `vaultWidgetAdapter.test.ts` covers only the conversion that still happens.

## Verdict

REQUEST_CHANGES

---

## How This Was Reviewed
5 review lenses over the 85915ecb..72beddb5 delta, each finding independently verified by codex. 1 candidate finding was refuted and dropped.

murderteeth:
> * [HIGH] Transactions slower than 3 minutes now dead-end

This one is new and looked worth flagging @w84april. I left the lows in too. I verified all these except `Genuine Safe failures show the wrong message`

murderteeth (dismissed):
lgtm

Reviewed the `72beddb5..d486fe8e` delta: the 3-minute timeout dead-end is fixed, the Speed-up/replacement tracking works as specified, and typecheck, boundary check, and all 1,021 tests pass with both preview deploys serving correctly.

Four medium follow-ups are logged as issues rather than blocking here:

- #1376 — planned-path receipt wait rejects a successful tx as a self-"repriced" replacement on load-balanced RPCs
- #1377 — connector `eth_getTransactionByHash` errors reject or stall the receipt wait
- #1378 — `vercel-deploy.yml` pins two different `yearn-gha` SHAs
- #1379 — clarify the production deploy path (root `vercel.json` / git-integrated projects)

---

## How This Was Reviewed
5 review lenses over the `72beddb5..d486fe8e` delta, each finding independently verified by codex. 1 candidate finding was refuted and dropped.

rossgalloway (dismissed):
LGTM

murderteeth (approved):
## Summary
Round 5 of this PR, reviewed against `origin/main`. The rebase is clean: the tree delta since the last review is 15 files, and 5 of the 6 substantive ones are byte-identical to `origin/main` — they arrived via merged PRs #1367 and #1369, not via new work here. Spot-checked the funds-sensitive moves (`stakingAdapter`, `useDirect*`, `approvalReset`, `approvalSpender`, `safeDepositBatch`, `safeWithdrawBatch`, both `valuation.ts`, `withdrawStepHelpers`, `slippage`, `ensoRouters`): every one is byte-identical to `main` apart from import paths. The `chainID` analytics property, the Base 2-confirmation rule, the 10% gas buffer, and the yBOLD security headers are all confirmed in place, the last one live in served responses. Typecheck, boundary check, and all 1,036 tests pass across the three workspaces; the yBOLD preview renders with live data and no console errors.

## Dependencies
`@next/env`, `@wagmi/core`, `@safe-global/safe-apps-sdk`, `@tanstack/query-core`, and `dayjs` are newly *declared* in `apps/yearn/package.json` but were already resolved transitively on `main` at the identical versions (`16.3.0`, `2.22.1`, `9.1.0`, `5.101.4`, `1.11.13`). No package version enters the tree that was not already there. No advisories.

## Issues
None blocking. One finding is tracked in #1382 (item 1): `ensoRoute.test.ts` has no counterpart at HEAD, so `normalizeEnsoRouteResponse` and `routeHasSwapStep` ship without coverage. The implementation is unchanged and nothing is broken today — no action needed on this PR.

## Verdict
APPROVE

---

## How This Was Reviewed
2 review lenses, each finding independently verified by claude. 1 candidate finding was refuted and dropped. Tier was `light` by argument; this review is not exhaustive at 2 lenses over a 1,135-file diff.

PR #1352, opened 2026-08-13
Feat/cache ledger

w84april:
## Description

Adds a ledger-backed portfolio data path that reuses a wallet’s indexed event history across balance, protocol-return, and per-vault growth calculations.

The portfolio now requests these results from one combined endpoint:

```text
GET /api/holdings/ledger/portfolio
```

The existing holdings implementation remains available as a fallback and has not been removed (yet).

Key changes:

- Store each wallet’s canonical Envio event history as one compressed Redis value.
- Incrementally refresh warm ledgers using per-chain coverage cursors and overlap windows.
- Batch the six event streams for each chain into one GraphQL request.
- Calculate balance history, protocol return, and vault growth concurrently from the same event revision.
- Share request-scoped vault metadata, PPS, and historical-price work across calculations.
- Cache daily USD balance totals for 30 days and protocol-return/growth results for 30 minutes.
- Add targeted invalidation for newly indexed or historically backfilled vaults.
- Add source and valuation revision markers for broader cache invalidation.
- Avoid persisting historical PPS and token-price timelines in Redis.
- Add a sortable Growth column to portfolio vault rows, including underlying growth, percentage return, and annualized-return tooltip.
- Add correlated debug timings and repeatable cold, warm, hot, repair, `1y`, and `all` benchmark tooling.
- Prefer `_PORTFOLIO_DEV` Upstash credentials when present, falling back to the existing portfolio Redis variables.

## Motivation and Context

The existing portfolio flow independently rebuilds historical context for balance and protocol-return requests. This repeats Envio event loading and duplicates historical PPS and price work, making portfolio charts expensive even when the wallet’s history has not changed. 
This PR also adds Growth column in portfolio.

The new path separates durable wallet history from derived valuations:

1. Envio events are synchronized into a reusable wallet ledger.
2. The server prepares one shared event and metadata context.
3. Balance history, protocol return, and per-vault growth run concurrently.
4. Revision-fenced derived results are cached and reused while their event, invalidation, valuation, and settled-day identities remain valid.

A fresh verified cache can return without loading the full wallet ledger or contacting Envio, Kong, or the historical-price provider.

## How to Review

Suggested review order:

1. `src/server/lib/holdings/services/ledger/walletSync.ts`
   - Cold bootstrap, warm overlap refresh, reconciliation, locking, and targeted backfill handling.

2. `src/server/lib/holdings/services/ledger/walletStore.ts`
   - Single-value compressed ledger format, verification, atomic updates, and freshness header.

3. `src/server/holdings/ledger/portfolio.ts`
   - Combined endpoint scheduling and concurrent balance, protocol-return, and growth calculations.

4. `src/server/lib/holdings/services/valuationLoader.ts`
   - Request-scoped PPS and historical-price batching, deduplication, and concurrency limits.

5. `walletTotalsCache.ts`, `walletDerivedCache.ts`, and `walletInvalidation.ts`
   - Revision fencing, daily USD caching, derived-result caching, and targeted invalidation.

6. Portfolio hooks and vault list components
   - Combined endpoint integration, legacy fallback, Growth column, sorting, formatting, and tooltip behavior.

The earlier snapshot-based ledger routes remain available for isolated testing, but they are not the primary portfolio client path.

## How Has This Been Tested?

- [x] TypeScript validation with `bun run tslint`
- [x] Biome formatting and lint validation with `bun run lint:fix`
- [x] Production build with `bun run build`
- [x] Focused holdings and portfolio test suites covering 351 tests
- [x] Cold, warm, and hot requests tested against development Upstash
- [x] Tested with small and large event-history wallets
- [x] Tested `1y` and `all` timeframe transitions
- [x] Tested unchanged refreshes, overlap repair, reconciliation, invalidation, concurrent requests, corrupt data, and upstream failures
- [x] Verified historical PPS and token-price timelines are not persisted in Redis
- [x] Verified incomplete upstream valuations are exposed as estimated instead of being cached as complete results

Manual smoke test:

1. Configure development portfolio Redis credentials and set `HOLDINGS_LEDGER_MODE=shadow`.
2. Open a populated wallet on `/portfolio`.
3. Confirm balance and protocol-return charts load.
4. Confirm each supported vault row displays Growth with two decimal places.
5. Sort the vault list by Growth and hover the dotted value to inspect total and annualized return.
6. Switch between `1y` and `all`.
7. Reload within 25 minutes and confirm the combined response uses the hot cache.
8. Set `HOLDINGS_LEDGER_MODE=off` and confirm the portfolio falls back to the legacy requests.

## Risk / Impact

The primary risk is portfolio data correctness around event reconciliation, historical valuation, and cache invalidation. The implementation mitigates this with checksummed ledger payloads, atomic revision-fenced writes, authoritative overlap replacement, periodic full reconciliation, targeted vault invalidations, and extensive regression coverage.

Other operational considerations:

- Production use requires `HOLDINGS_LEDGER_MODE=read-write`.
- The default mode remains `off`.
- `shadow` mode supports development and rollout validation.
- The existing balance and protocol-return implementation remains as a client fallback.
- No transaction, deposit, withdrawal, or funds-handling code is changed.
- Historical PPS and price timelines are kept in request memory rather than permanently stored.
- Indexer repairs can be applied through targeted vault invalidation.
- `HOLDINGS_LEDGER_SOURCE_REVISION` forces an event-source rebuild when required.
- `HOLDINGS_LEDGER_VALUATION_REVISION` invalidates derived valuations without replaying events.
- Rollback is configuration-only: set `HOLDINGS_LEDGER_MODE=off`.


rossgalloway:
Quick review:

image
image

- Growth values should adhere to existing numerical formatting that exists elsewhere on the site. Generally, it holds values in lists to 3 digits. The value in the image above would be shortened to `+1.47K` instead of `+1,476.09`. This will keep the column looking less cluttered.
- I think we should be consistent with units in the vault rows here and convert everything to dollars. We show all other values in dollars in the row and it makes sense to do the same with growth. We can then show a USDC growth value in the expanded view of the row, either as a separate value or in the existing hover in the "Your Position" chart.
- locked yvUSD growth values are not being shown in the row. Not sure why not, as they do show in the "Your position" chart on expand and the Vault Performance charts above.

w84april:
Experimental PR, moved good stuff here https://github.com/yearn/yearn.fi/pull/1385

PR #1351, opened 2026-08-12
Fix and track Enso cross-chain transactions

rossgalloway:
### Summary
Cross-chain Enso vault actions could stall after approval, report success before bridge delivery, or keep polling without a clear result. This change separates calibration from executable quotes, checks the final transaction before submission, and tracks bridge delivery through Relay, Stargate, or CCIP.

### How to review
Start with the protected quote flow in `packages/vault-widget/src/components/widget/shared/useProtectedEnsoQuoteState.ts`. Confirm that a cross-chain calibration request may use one basis point, but its transaction can never execute.

Then review raw Enso execution in `packages/vault-widget/src/hooks/useEnsoOrder.ts` and bridge tracking in `apps/yearn/src/components/shared/hooks/useEnsoBridgeStatusPoller.ts`. The host app owns persisted notifications, bridge polling, and targeted source and destination balance refreshes.

Test a cross-chain deposit or withdrawal that needs approval. Confirm that the widget advances after approval, reports a failed source transaction, stays in progress after source confirmation, and reports success only after destination delivery.

### Test plan
- [ ] Manual: Complete Ethereum to Base and Ethereum to Katana vault routes with a wallet. Check approval handoff, source failure handling, bridge progress, destination transaction links, and updated balances.
- [x] Automated: `bun run tslint`
- [x] Automated: Vault widget suite, 34 files and 252 tests
- [x] Automated: Yearn app suite, 157 files and 781 tests
- [x] Automated: `bun run check:vault-widget-boundary`
- [x] Automated: `bun run build`

### Risk / impact
This changes transaction execution and status reporting for vault deposits and withdrawals, including fund-moving cross-chain routes. The final executable quote never exceeds the user slippage setting. Enso bridge checks run through the server with `ENSO_API_KEY`, and one app-level poller limits requests to one every ten seconds.

If automatic tracking is unavailable, the notification stays submitted and directs the user to the source transaction instead of reporting false success. Reverting this PR restores the previous quote and notification behavior.

PR #1354, opened 2026-08-13
Force Enso vault zaps through router

rossgalloway:
### Summary
Vault zaps could receive an Enso delegate-call route when a smart account was connected outside the Safe app. The widget cannot execute delegate calls, so the quote was blocked by the router safety check.

This change makes router execution an invariant inside the vault deposit and withdrawal hooks. It also rejects delegate-call responses before they enter widget state.

Stacked on #1351.

### How to review
Review the route boundary first:

1. Check useEnsoDeposit and useEnsoWithdraw. Both always pass routingStrategy=router.
2. Check ensoRoute. Any non-zero or unknown operationType is rejected before the route is exposed.
3. Check the removed widget and flow props. Safe detection still controls Safe batching and tracking, but no longer controls Enso routing.
4. Check the pure request serializer and focused regression tests.

### Test plan
- [x] Automated: bun run test:all (1,071 tests passed)
- [x] Automated: bun run tslint:all
- [x] Automated: bun run check:vault-widget-boundary
- [x] Automated: Biome check on all 12 changed files
- [x] Automated: bun run build
- [x] Manual: reproduced the reported smart-account USDS vault to USDC route with routingStrategy=router; Enso returned the documented router and no delegate operation

### Risk / impact
This affects transaction preparation for all Enso vault deposits and withdrawals. It reduces the supported route set to the normal router execution mode that the widget already approves, simulates, batches, and validates.

The existing router allowlist remains active. Delegate transactions never reach approval or execution. Reverting commit 23a078d7 restores the previous strategy selection.

### Note
A version of this fix is also in #1355 for the main branch

### DIFF
--- a/packages/vault-widget/src/components/widget/deposit/index.tsx
+++ b/packages/vault-widget/src/components/widget/deposit/index.tsx
@@ -385,7 +385,6 @@ export function WidgetDeposit({
     slippage: ensoQuoteSlippage,
     ensoQuotePurpose: ensoQuoteRequest.purpose,
     ensoEnabled,
-    ensoRoutingStrategy: isWalletSafe ? 'router' : undefined,
     routeRefreshKey: approvalRouteRefreshKey,
     stakingSource
   })
--- a/packages/vault-widget/src/components/widget/deposit/useDepositFlow.ts
+++ b/packages/vault-widget/src/components/widget/deposit/useDepositFlow.ts
@@ -4,7 +4,7 @@ import { useDirectStake } from '@yearn/vault-widget/internal/hooks/actions/useDi
 import { useEnsoDeposit } from '@yearn/vault-widget/internal/hooks/actions/useEnsoDeposit'
 import { useYBoldZapperDeposit } from '@yearn/vault-widget/internal/hooks/actions/useYBoldZapperDeposit'
 import { useYvUsdLockedZapDeposit } from '@yearn/vault-widget/internal/hooks/actions/useYvUsdLockedZapDeposit'
-import type { EnsoQuotePurpose, EnsoRoutingStrategy } from '@yearn/vault-widget/internal/hooks/solvers/useSolverEnso'
+import type { EnsoQuotePurpose } from '@yearn/vault-widget/internal/hooks/solvers/useSolverEnso'
 import { toAddress } from '@yearn/vault-widget/internal/utils'
 import { toBasisPoints } from '@yearn/vault-widget/internal/utils/slippage'
 import { YVUSD_LOCKED_ADDRESS } from '@yearn/vault-widget/internal/utils/yvUsd'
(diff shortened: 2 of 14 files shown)


rossgalloway (commented):
## Summary

The vault action hooks correctly centralize router-only Enso execution, and route normalization prevents explicit delegate transactions from reaching approval, simulation, or execution consumers.

However, one reachable Enso request path remains outside this invariant: the native-token MAX quote flow.

## Issues

- `packages/vault-widget/src/components/widget/deposit/useFetchMaxQuote.ts:95` — The native-token MAX flow constructs its Enso request directly and omits `routingStrategy=router`. For an affected smart account, Enso can therefore return a delegate route. The new normalizer rejects that response, both MAX quote attempts can return without a route, and `onResult` is never called.

  Build this request with `buildEnsoRouteRequestParams` and `VAULT_ENSO_ROUTING_STRATEGY`, preserving the existing two-pass slippage and gas-reserve behavior. Add same-chain and cross-chain regression coverage proving that MAX requests include `routingStrategy=router`.

## Suggestions

- Add an interaction-level smart-account check covering quote readiness and transaction-plan creation for both deposit and withdrawal. Screenshot comparison is unnecessary, but checking only the raw Enso response does not prove that the widget exposes an executable plan or that Safe batching remains intact.

## Verdict

COMMENT

---

## How This Was Reviewed

Reviewed with the [review-pr-workflow skill](https://github.com/yearn/webops-skills/tree/main/skills/review-pr-workflow) — 5 review lenses, with each material finding independently verified by a same-model agent in an independent context. 0 candidate findings were refuted and dropped.

w84april (dismissed):

murderteeth (dismissed):
## Summary

Router-only execution is enforced at the two hooks that matter, the shared serializer removes the duplicated URL construction, and the delegate rejection sits at the single normalization path both request routes go through. The prior review's `useFetchMaxQuote` finding is genuinely fixed by `b1e206e4`. No blocking defects survived verification.

## Verdict

APPROVE

---

## How This Was Reviewed
5 review lenses, each finding independently verified by claude. 1 candidate finding was refuted and dropped.

---

`tier: full — sensitivity trigger (on-chain transaction preparation, routing-strategy selection, delegate-call rejection ahead of a router allowlist).` `verify-agent: claude. confirmed: 0. refuted: 1. advisories: 0. unverified past cap: 0.`

murderteeth (approved):
## Summary
Router execution is now an invariant at both ends: the two action hooks pass `VAULT_ENSO_ROUTING_STRATEGY` unconditionally, and `normalizeEnsoRouteResponse` drops any route whose `tx.operationType` is present and non-zero before it can reach approval, simulation, or the Safe batch. `buildEnsoRouteRequestParams` reproduces the previous cross-chain serialization exactly — `destinationChainId` is still emitted only when it differs from the source chain — so the extraction is behavior-preserving for both the solver and the MAX-quote path. Removing `ensoRoutingStrategy` from the widget and flow props leaves `isWalletSafe` doing only what it should: Safe batching and tracking.

## Issues
None. No defects found.

Verified locally on the PR head against `origin/codex/enso-tracking-monorepo`: `biome check packages/vault-widget/src` clean, `bun run tslint:widget` clean, `bun run test:widget` 304/304 passing.

## Verdict
APPROVE

---

## How This Was Reviewed
5 review lenses, each finding independently verified by claude. 1 candidate finding was refuted and dropped.

PR #1364, opened 2026-08-17
Feat/ledger

w84april:
### Summary

Replaces the durable portfolio-ledger prototype from #1352 with a simpler combined portfolio endpoint built around the existing holdings calculators. The new architecture preserves balance and protocol-return semantics while sharing request context, adding focused cache layers, and substantially improving cold-load performance.

This also improves portfolio Growth reporting: vault rows and the USD Growth chart now value asset-denominated yield using the latest price already fetched by the API, while receipt-weighted values remain available for return percentages and indexes.

### Highlights

- Add a combined `GET /api/holdings/portfolio` endpoint with client-side legacy fallback.
- Share wallet events, metadata, PPS, and price work between balance and protocol-return calculations.
- Add a short-lived compressed wallet-event cache, date-partitioned balance totals, and settled protocol-return snapshots with guarded tail appends.
- Use bounded Envio pagination, batched/indexed lookups, and process-wide price-provider concurrency.
- Normalize Kong decimal metadata and safely repair compatible cached responses.
- Calculate sortable per-vault USD growth and combine related yvUSD and yBOLD positions.
- Allow a later valid asset price to recover Growth rows when a receipt-time price is missing.
- Keep receipt-weighted `growthWeightUsd` for protocol-return percentages and indexes.
- Add a USD contribution chart containing the four largest vault contributions, `Other`, and the authoritative portfolio total.
- Keep named contributions plus `Other` equal to total Growth at every chart point.
- Align desktop column headers, show TVL beside Growth on mobile, compact sub-cent values, and use neutral Growth colors.

### Performance

Measured using a production build and the heavy wallet `0x93A62dA5a14C80f265DAbC077fCEE437B1a0Efde`:

| Scenario | Result |
| --- | ---: |
| Original correct sequential implementation | 70.48s |
| Simplified cold request | 21.85s |
| Immediate warm median | 1.65s |
| Seven-day repair | 22.67s |
| Previous ledger controlled range | 18.86–23.61s |

Cold, warm, event-cache-miss, and repaired responses were verified for parity as documented in `docs/performance/holdings-portfolio-simplification-2026-08-16.md`.

### How to review

1. Start with `src/server/lib/holdings/services/portfolio.ts` and the new API route to review combined request orchestration.
2. Review `settledHoldingsContext.ts`, `walletEventCache.ts`, and `cache.ts` for sharing, cache identity, invalidation, and tail behavior.
3. Review `pnlSimple.ts` and `protocolReturnFamilySeries.ts` for:
   - latest-price `growthUsd`;
   - receipt-weighted return preservation;
   - missing receipt-price recovery;
   - yvUSD/yBOLD aggregation;
   - family-series ranking and compaction.
4. Review `usePortfolioHistoryBundle.ts`, `PortfolioHistoryChart.tsx`, and `PortfolioGrowthContributionsChart.tsx` for the client data flow and contribution visualization.
5. Review the vault-list changes for responsive layout and Growth formatting.

### Test plan

- [x] `bun run tslint`
- [x] `bun run lint:fix`
- [x] `bun run test` — 182 files, 986 tests
- [x] Verify cold, warm, event-cache-miss, and seven-day repair response parity
- [x] Verify the heavy-wallet balance history contains no carried-price anomaly
- [x] Verify yvUSD and yBOLD single/combined position Growth
- [x] Verify missing receipt prices do not blank otherwise calculable Growth rows
- [x] Verify the contribution stack conserves total Growth at every point
- [x] Verify mixed positive/negative contributions, sparse history, timeframe rebasing, and stable ranking
- [x] Verify mobile TVL placement, aligned headers, compact sub-cent values, and neutral Growth colors

### Risk / impact

- This is a read-only portfolio-data change; it does not submit transactions or modify funds.
- Cache-key versions are updated, so the first request after deployment will perform a cold recomputation.
- Adding per-family `growthUsd` increases the uncompressed history payload, although the additional compressed transfer size is comparatively small.
- The optional Kong batch-PPS path remains disabled by default and can be turned off independently.
- The client retains its legacy endpoint fallback, providing a straightforward rollback path for the combined route.


rossgalloway (commented):
## Summary

The branch typechecks, builds, passes 145 focused tests, and renders correctly in desktop/mobile browser checks; root Biome remains blocked by 221 untouched SVG parse errors.

## Issues

[ ] src/server/lib/holdings/services/pnlSimple.ts:363 — Nested-vault performance undercount (high) — Locked yvUSD uses only its outer PPS, denominated in yvUSD, so both Index and Position modes omit the underlying yvUSD-to-USDC PPS growth.
- Change: Calculate terminal-underlying PPS recursively; for Locked yvUSD, use `lockedPPS × yvUSDPPS`.
- Keep: Preserve separate locked/unlocked series, position-size-neutral Index behavior, and receipt-based accounting.
- Done when: Tests prove both modes include both PPS layers and report Locked yvUSD growth in USDC terms.

- [ ] **src/server/lib/holdings/services/kong.ts:72 — Preview credential disclosure (high)** — the Vercel bypass secret is sent to any configured Kong/PPS origin and can follow cross-origin redirects.
  - Change: Send the header only to an exact HTTPS allowlisted preview origin and reject or manually handle cross-origin redirects.
  - Keep: Preserve server overrides, batch PPS support, and content-type behavior.
  - Done when: Tests prove default, unrelated, HTTP, and redirected origins never receive the secret.

- [ ] **src/server/lib/holdings/services/pnlSimple.ts:2520 — Same-day Growth cutoff drift (medium)** — the new Growth snapshot can include activity after the chart’s latest settled date, then freeze that timing-dependent value in the daily cache.
  - Change: Build final Growth ledgers from events at or before `latestTimestamp`.
  - Keep: Preserve settled history, same-transaction enrichment, and the shared event cache.
  - Done when: Post-cutoff events affect neither Growth nor its cached snapshot until their settled day.

- [ ] **src/server/lib/holdings/services/portfolio.ts:57 — Nested PPS invalidation gap (medium)** — a warm combined response can retain stale historical balances when only a nested vault dependency is invalidated.
  - Change: Validate the balance cache against the complete top-level and nested PPS dependency set.
  - Keep: Fully valid warm hits must still avoid loading wallet events and valuation context.
  - Done when: Invalidating only a nested vault forces the affected totals to recalculate.

- [ ] **src/server/lib/holdings/services/walletEventCache.ts:154 — Unreadable cache writes (medium)** — oversized event payloads can be saved successfully even though the reader must reject them, causing repeated full refetch/compress/write cycles.
  - Change: Enforce both encoded and decoded limits before `redis.set`, or use a bounded chunked format.
  - Keep: Preserve Brotli, schema validation, hashed keys, TTL, and fail-open behavior.
  - Done when: Every saved value is readable and boundary tests reject over-limit writes.

## Suggestions

- **Locked yvUSD growth calculation issues** — Implement generic nested-vault PPS composition rather than an address-specific Locked yvUSD exception.
- **Production Redis selection (high)** — Consider allowing `_PORTFOLIO_DEV` credentials only behind an explicit preview/development signal and namespace shared keys by environment.
- **Fallback integration coverage (medium)** — Mount the coordinator and test schema errors, wallet changes, refresh failures, and combined-to-legacy handoff.
- **Endpoint parity coverage (medium)** — Add combined-versus-legacy tests across USD/ETH, `1y`/`all`, cold/warm/tail, and vault-version scopes.
- **Provider/load coverage (medium)** — Exercise real Envio, Prices, Kong, and Upstash behavior for maximum-history wallets and concurrent instances.
- **Deployment-matrix coverage (medium)** — Validate Redis precedence, Kong URL normalization, batch PPS on/off, and protected-preview behavior in each runtime.
- **Growth interaction coverage (low)** — Browser QA covered desktop/mobile layout, sorting, and tooltip behavior; keyboard and uncommon negative/error states remain untested.
- **Component boundaries (low)** — Consider extracting portfolio Growth into a declarative metric cell so the shared vault row does not own portfolio-specific state and copy.
- **Service signatures (low)** — Consider named options for the combined orchestration APIs to remove positional `undefined` placeholders.

## Verdict

COMMENT

---

## How This Was Reviewed

5 review lenses, each material finding independently verified by `session` using fresh same-model contexts. 0 candidate findings were refuted and dropped.

rossgalloway:
UX review:

Much faster and snappier loading! That is great!

Growth value isn't showing up for Katana vbUSDC vault: 
image

The green growth text color is distracting to me. Everything else is white and this is green. Feel out of place. Green also implies it could be red, which we very much do not want.

Not a fan of the "Worst Performance" text in the Vault Performance dropdown: 
image

Overall, I find the Vault Performance chart to often be more confusing than illuminating. I'm wondering if we can somehow combine it with the growth chart and have the proportion of area under the curve represented by the vaults that contributed. Will mock something up.

rossgalloway:
What about something like this: 
image

vault growth is shown within the overall growth chart. I find this much easier to understand. And can work for all the same basis values as the growth chart

rossgalloway (changes requested):
### Locked yvUSD still not showing correct values

Locked yvUSD is still using the unlocked yvUSD vault as it's assumed underlying asset and is undervaluing the growth from deposits. The growth value needs to price locked yvUSD as `lockedYvUSDPPS * unlockedYvUsdPPS * USDC price`and then make sure if the user has both positions that they are blended correctly.

What it shows: 

Image

What it should show: 

Image

### Add more vaults to constituent growth chart

 I also recommend increasing the number of vaults that can be shown in the growth constituent chart from 4-8 (or more). For 0x16388463d60FFE0661Cf7F1f31a7D658aC790ff7 compare:  

Image

with

Image

### Include constituents in ETH and INDEX versions of growth chart

This data is available and can be added just like with the USD values.

### Remove Vault Performance chart

I think we can remove the vault performance tab now the information that was contained there is now in the Growth chart.

w84april:
Closed in favour of https://github.com/yearn/yearn.fi/pull/1385

PR #1366, opened 2026-08-19
Add site status indicator

rossgalloway:
### Summary
Per request for Engn33r, AIO guidelines recommend a visible last updated element on the site. This adds a responsive status indicator with the build time, Kong health, and supported RPC availability. The status element lives in the bottom right of the screen when in desktop mode with space on the side, and below the wallet button if there is not available space beside the content. On mobile it is in the hamburger menu

### How to review
Start with `src/server/status.ts` and `app/api/status/route.ts` to review the health checks and cache policy. Then review `src/components/shared/components/SiteStatus.tsx` for the responsive display.

Check these layouts:
- Standard desktop: the summary appears below the header and expands on hover.
- Wide desktop at 1700px or more: the full status appears vertically in the bottom-right corner.
- Mobile: the full date and service details appear on two centered lines at the bottom of the navigation menu.
- Vault welcome prompts remain above the wide desktop status.

### Test plan
- [x] Manual: verified standard desktop, hover expansion, wide desktop, mobile navigation, and welcome prompt positioning with Playwright.
- [x] Manual: verified the live status route reports Kong and all 8 supported RPCs without exposing endpoint URLs.
- [x] Automated: `bun run test` (923 tests)
- [x] Automated: production build

### Risk / impact
This does not affect funds, wallet connections, authentication, or database state. The public route makes read-only checks to Kong and 8 configured RPC endpoints in parallel, uses a 4-second timeout, and is cached at the Vercel CDN for 30 seconds. The client refreshes once per minute.

The change can be rolled back by reverting the feature commit.

### DIFF
--- a/apps/yearn/app/api/status/route.ts
+++ b/apps/yearn/app/api/status/route.ts
@@ -1,0 +1,4 @@
+export const runtime = 'nodejs'
+export const dynamic = 'force-dynamic'
+
+export { GET, OPTIONS } from '@/server/status'
--- a/apps/yearn/next.config.ts
+++ b/apps/yearn/next.config.ts
@@ -4,6 +4,7 @@ import type { NextConfig } from 'next'
 
 const WORKSPACE_ROOT = fileURLToPath(new URL('../..', import.meta.url))
 loadEnvConfig(WORKSPACE_ROOT, process.env.NODE_ENV !== 'production', console, true)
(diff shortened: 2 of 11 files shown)


rossgalloway:
going to change this so this data lives in the headers and console, and in the setting menu. if this is for AIO, they will get the headers and can easily read console data. We don't need to pollute the main page with it.

PR #1367, opened 2026-08-19
Fix vault strategy allocation display issue

rossgalloway:
### Summary
Anonymous users (users who do not have a wallet connected) receive a compact server-rendered vault list. This list did not have enough data to populate the vault strategy allocation chart in the expanded vault list sections, so they were left blank.

Before: image

This change fills the missing strategy composition and total assets from the snapshot.

After: 
image

### How to review
Start with `kongVaultSelectors.ts` and confirm that compact vault views use snapshot strategies only when their strategy list is empty.

Then review the selector tests and the anonymous expanded-row regression. To test manually, open `/vaults` without connecting a wallet, expand yvUSD, and confirm that the strategy allocation is visible.

### Test plan
- [x] Manual: Expanded yvUSD from the anonymous vault list and confirmed that allocated strategies render without console errors or failed requests.
- [x] Automated: `bun run lint`
- [x] Automated: `bun run test` (925 tests)
- [x] Automated: Focused selector and expanded-row tests (25 tests)

### Risk / impact
Low risk. This changes read-only vault display data and does not change deposits, withdrawals, transactions, or wallet behavior. Existing non-empty strategy lists remain unchanged, and explicit zero TVL values stay zero.

Rollback is a revert of this commit.

### DIFF
--- a/src/components/pages/vaults/components/list/VaultsListRowExpandedContent.test.tsx
+++ b/src/components/pages/vaults/components/list/VaultsListRowExpandedContent.test.tsx
@@ -1,0 +1,105 @@
+import type { TKongVaultInput } from '@pages/vaults/domain/kongVaultSelectors'
+import { renderToStaticMarkup } from 'react-dom/server'
+import { describe, expect, it, vi } from 'vitest'
+import VaultsListRowExpandedContent from './VaultsListRowExpandedContent'
+
+const { snapshot } = vi.hoisted(() => ({
+  snapshot: {
+    address: '0x1111111111111111111111111111111111111111',
+    chainId: 1,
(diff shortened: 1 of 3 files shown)


murderteeth (approved):
## Summary
Fills strategy composition and total assets from the Kong snapshot so anonymous SSR vault views can render the allocation chart. Verified against live Kong data: the anonymous /vaults expanded row renders the donut, and the derived asset price reconciles on both a stablecoin vault (yvUSD, $8.84M vs. $8.84M TVL) and an ETH-priced one (WETH-1, $19.6M vs. $19.6M).

## Issues
None.

## Verdict
APPROVE

PR #1369, opened 2026-08-20
Use staked net APY for yBOLD estimates

rossgalloway:
### Summary
Per discussion with Strategists, yBOLD should use the higher of the staked 7-day net APY and staked Oracle net APY for the Est. APY in the interface, page metadata, and structured data. This change makes the Est. APY more faithful to the true rate.
image
The grey oracle line under-estimates significantly. The 7 day value is what is currently used, but but can drop below the oracle rate, which should rarely happen. 

This PR also fixes an issue with page hydration showing a gross APY estimated value for yBOLD momentarily before reverting to the netAPY value.

### How to review
Start with the yBOLD calculation in `vaultApy.ts` and the staked data selection in `kongVaultSelectors.ts`. Then review the related snapshot prefetch in `publicDataHydration.ts` and the server metadata handling in `app/metadata.ts`.

To check the main flow, hard refresh the vanilla yBOLD vault page. The estimated APY should remain stable during hydration, and the metadata description and JSON-LD value should match the visible APY.

### Test plan
- [x] Manual: Hard refreshed the yBOLD vault page and confirmed the visible APY remains 4.95% before and after hydration
- [x] Manual: Confirmed the metadata description and JSON-LD annual percentage rate also report 4.95%
- [x] Automated: `bun run lint`
- [x] Automated: `bun run test` (175 files, 928 tests)

### Risk / impact
This changes displayed yield data and metadata only. It does not change deposits, withdrawals, transaction construction, or other funds-moving behavior.

The vanilla yBOLD metadata path makes one additional request for the staked snapshot. If either required snapshot is unavailable, the page omits the yBOLD-specific metadata instead of publishing the incorrect vanilla yield. Reverting this PR restores the previous behavior.

### DIFF
--- a/app/metadata.test.ts
+++ b/app/metadata.test.ts
@@ -1,0 +1,92 @@
+import { YBOLD_STAKING_ADDRESS, YBOLD_VAULT_ADDRESS } from '@pages/vaults/domain/yBoldProduct'
+import { buildVaultSnapshotEndpoint } from '@shared/data/publicQueryEndpoints'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { buildVaultMetadataFromInput, buildVaultStructuredDataFromInput, fetchVaultMetadataSnapshot } from './metadata'
+
+const { fetchWithSchemaMock } = vi.hoisted(() => ({
+  fetchWithSchemaMock: vi.fn()
+}))
+
+vi.mock('@shared/utils/fetchQuery', () => ({
+  fetchWithSchema: fetchWithSchemaMock
+}))
+
+const BASE_SNAPSHOT = {
+  meta: {
+    displayName: 'yBOLD',
+    displaySymbol: 'yBOLD',
+    token: { symbol: 'BOLD' }
+  },
+  apy: {
+    net: 0.0552,
+    weeklyNet: 0.0552
+  },
+  performance: {
+    estimated: { apy: 0.2 },
+    oracle: { netAPY: 0.0552 }
+  },
+  tvl: { close: 5_890_000 }
+} as any
+
+const STAKED_SNAPSHOT = {
+  apy: {
+    net: 0.0495,
+    weeklyNet: 0.046
+  },
+  performance: {
+    estimated: { apy: 0.2 },
+    oracle: { netAPY: 0.0495 }
+  }
+} as any
+
+describe('yBOLD vault metadata', () => {
+  beforeEach(() => {
+    fetchWithSchemaMock.mockReset()
+  })
+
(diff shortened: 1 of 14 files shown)


rossgalloway:
Additional analysis with backtesting showing how the new Estimated value compares to the old.

New:
image

Old: 
image

murderteeth (commented):
## Summary
yBOLD's Est. APY becomes the higher of the staked 7-day net APY and the staked Oracle net APY, across the vault list, detail header, page description and JSON-LD. Verified live against the preview deploy: list row, header, `<meta name="description">` and `annualPercentageRate` all report **5.87%** (staked oracle net), versus production's 5.53% shown / 6.54% published — the parity claim in the PR body holds, and the hydration flash is gone (production still shows a 0.00% frame at ~1.2s; the preview is stable from first paint).

## Issues
- [ ] **app/metadata.ts:95** — **Metadata blanks out on a partial fetch failure (medium)** — if the newly added staked-snapshot request fails while the vault's own request succeeds, the page drops its real title, symbol, description and TVL and falls back to the generic "Yearn Vault" placeholders. With `revalidate = 21600` on both vault routes, one transient blip is served that way for up to six hours. The PR body describes this as omitting "the yBOLD-specific metadata" — the blast radius is wider than that.
  - Done when: with the vanilla snapshot succeeding and the staked one rejecting or timing out, chain 1 / `0x9F43…3d8` still publishes the yBOLD name, symbol and TVL and only the APY figure is absent, pinned by a test in `app/metadata.test.ts`.
  - Provenance: 862d1d71

- [ ] **src/components/pages/vaults/domain/kongVaultSelectors.ts:633** — **Degraded path shows the unstaked rate (medium)** — the unmerged base snapshot now takes precedence over the staked-merged list vault, so if the staked snapshot is missing the header renders the vault-level oracle net APY (6.54% on live data) instead of the staked 5.87% — an over-estimate where `main` rendered 0.00%.
  - Done when: on a hard refresh of the vanilla yBOLD page with the staked snapshot request failing, the Est. APY is never the unstaked oracle rate.
  - Provenance: 7370323a

## Verdict
COMMENT

---

## How This Was Reviewed
5 review lenses, each finding independently verified by claude. 3 candidate findings were refuted and dropped.

murderteeth (approved):
## Summary
Verified against the preview deploy and live Kong data. The vanilla yBOLD route now publishes and renders the same number — 6.16% in the list row, detail header, meta description and JSON-LD — per the spec. Production renders 5.68% while publishing 6.87%, so the parity fix is real, and the hydration flash is gone. Both issues from the previous round check out.

## Issues
None.

## Verdict
APPROVE

PR #1385, opened 2026-08-31
Feat/ledger monorepo

w84april:
### Summary

This PR adds per-vault Growth details to the existing Portfolio experience and updates the existing Growth charts to show which vaults generated returns.

It also replaces the page’s separate balance-history and protocol-return requests with one `GET /api/holdings/portfolio` request. The response contains balance history, protocol-return history, and per-vault Growth, allowing the calculations to run in parallel while sharing the same wallet event and metadata context.

### Portfolio changes

- Break down the existing Growth USD and ETH charts into contributions from up to eight vaults, with the remainder grouped as “Other” and the portfolio total shown alongside them.
- Compare the portfolio Growth index with up to eight individual vault indexes.
- Add a sortable Growth column to vault positions, with details for asset growth, Real APY, and total return.
- Identify approximate USD growth with an asterisk and explanation in chart and position tooltips.
- Combine locked and unlocked yvUSD Growth, and vault and staked yBOLD Growth, to match their existing grouped portfolio rows.
- Keep the previous Growth snapshot visible while a new combined history request is loading.
- Remove Vault Performance tab

### Server and data changes

- Add `GET /api/holdings/portfolio`, returning balance, protocol-return, and per-vault Growth in one response.
- Run balance and Growth calculations in parallel while sharing a lazily created wallet event and metadata context.
- Simplify Envio event fetching to one bounded pagination path that combines V2 and V3 activity.
- Make yearn-prices the only historical asset-price provider; Kong remains the source for vault metadata and price-per-share history.
- Add per-vault USD and underlying-asset Growth, including realized Growth valued at exit.
- Add a five-minute compressed wallet-event cache.
- Version-bump the daily-total and protocol-return/Growth caches, compress protocol-return snapshots, and extend their lifetime to 30 days.
- Skip long-lived cache writes when tracked metadata, PPS, or price requests fail.

### How to review

1. Review the combined contract and orchestration:
   - `apps/yearn/src/server/holdings/portfolio.ts`
   - `apps/yearn/src/server/lib/holdings/services/portfolio.ts`
   - `apps/yearn/src/components/pages/portfolio/hooks/usePortfolioHistoryBundle.ts`
2. Review the ledger and per-vault Growth calculations in `pnlSimple.ts` and its tests.
3. Review the new contribution and index visualizations, Growth mapping, and vault-list integration.
4. Review Envio, yearn-prices, and Kong request behavior.
5. Review Redis caching, invalidation, configuration, and public API compatibility.

### Test plan

- [ ] `bun run lint`
- [ ] `bun run tslint`
- [ ] `bun run test` — 901 tests passed, but Vitest exited non-zero because three workers failed during startup with `ERR_REQUIRE_ESM` before their tests were collected
- [ ] Smoke-tested `/portfolio` with the local Agent Wallet at desktop and mobile widths
- [ ] Confirmed the combined `/api/holdings/portfolio` request returned successfully
- [ ] Checked Balance, Growth, Annualized %, and Vault Performance
- [ ] Checked USD, ETH, Index, and Position views; 30D and 1Y ranges; chart tooltips; Top/Bottom contributor controls; and the position Growth tooltip
- [ ] Check 90D and ALL ranges
- [ ] Check holdings-table Growth sorting
- [ ] Check Redis cold loads, cache hits, and vault invalidation
- [ ] Add regression coverage for initial-request errors, no-activity wallets, nested-vault invalidation, and sub-dollar Growth ranges
- [ ] Run a clean production build

### Risk / impact

- Financial-data correctness is the main risk, especially transfer matching, staking families, receipt-time pricing, nested vaults, and incomplete upstream responses.
- The `/api/holdings/pnl/simple-history` compatibility route is removed. Confirm that no external consumers still depend on it before merging.
- `version`, `fetchType`, and `paginationMode` are no longer read by the affected holdings endpoints. Requests containing them still succeed, but the parameters are ignored and results combine V2 and V3 activity.
- Historical valuation now uses yearn-prices without the previous DefiLlama fallback. Deployments need `YEARN_PRICES_API_KEY` or `API_KEY_PORTFOLIO`.
- Redis remains optional. New cache-key versions begin cold, which may temporarily increase Envio, Kong, and yearn-prices traffic after deployment.
- No manual Redis migration is required. Old totals and snapshots expire naturally, while existing vault-invalidation markers remain shared.
- When both Redis credential pairs are configured, the `_PORTFOLIO_DEV` pair takes precedence regardless of `NODE_ENV`.
- No package manifest, lockfile, or database schema changes are included.
- Reverting requires no database rollback, but Redis entries written by this version remain until their configured TTLs expire.


rossgalloway (changes requested):
## Issues

  - [x] apps/yearn/src/server/lib/holdings/services/pnlSimple.ts — Support generic nested-vault valuation
      - Issue: While the fix here solves the problem for yvUSD specifically, we should expand the valuation logic to work for generic nested vaults. 
      - Change: Resolve each vault’s terminal asset recursively from vault metadata and compound PPS across every nested vault layer. Use terminal-asset units
        consistently for deposits, withdrawals, transfers, receipt prices, exits, and current-position valuation. If any required metadata or intermediate PPS
        is unavailable, report the valuation as incomplete instead of mixing intermediate share units with terminal-asset units.

      - Keep: Existing single-layer vault and staking-wrapper behavior must remain correct.
      - Done when: Regression tests cover multiple nested layers, transfer and deposit paths, current valuation, and missing intermediate PPS. Include a test
        proving that a three-layer position compounds all PPS values and that existing locked-vault valuation remains correct.

  - [ ] Portfolio Growth chart — Render Index as additive stacked attribution to match USD and ETH charts
      - Change: Provide per-vault additive index contributions through the server response, cache schema, and frontend types. Render Index using the same
        stacked attribution model as USD and ETH: a starting-index band of 100, up to eight named vault contributions, an exact Other residual, and the
        portfolio index as the total line. Independently rebased comparison lines do not provide additive portfolio attribution.

      - Keep: Existing timeframe rebasing, missing-data handling, USD/ETH views, and the eight-vault display limit.
      - Done when: At every chart date, 100 + named vault contributions + Other equals the displayed portfolio index, including after changing timeframe. Add
        tests covering positive and negative contributions and omitted vaults included in Other.

  - [ ] PortfolioGrowthContributionsChart.tsx — Reconcile signed stacks to the portfolio-total line
      - Change: Do not rely on stackOffset="sign" for mixed positive and negative contributions; it creates separate positive and negative stacks whose visible
        boundary does not equal the algebraic portfolio total. Build explicit cumulative range bands, or an equivalent representation, so all contribution bands
        terminate at the portfolio-total value.

      - Keep: Preserve the exact numeric Other residual, constituent tooltip rows, estimated-value indicators, and all contribution colors.
      - Done when: A regression case with +100 gain, −90 loss, and +10 portfolio growth produces a final stack boundary of +10, aligned with the total line. The
        Y-axis domain must include every intermediate band boundary.

rossgalloway (changes requested):
## Issues

- [ ] **Stale USD Growth pricing (high)** — [pnlSimple.ts:2359](/home/dev/raaaws/yearn.fi/apps/yearn/src/server/lib/holdings/services/pnlSimple.ts:2359): current per-vault USD Growth uses prices fetched around past receipts, so its dollar value can remain months out of date.
  Done when: current Growth uses the latest settled asset price while preserving receipt-weighted charts and exit-time realized values.
  Provenance: `dc3be54d`.

- [ ] **Missing Index contributions (medium)** — [pnlSimple.ts:2858](/home/dev/raaaws/yearn.fi/apps/yearn/src/server/lib/holdings/services/pnlSimple.ts:2858): equal vault gains and losses produce zero contribution increments, hiding both vaults’ performance despite the correct portfolio total.
  Done when: two $100 positions gaining and losing $25 show +12.5 and −12.5 Index contributions, with the total remaining 100.
  Provenance: `11919947`.

- [ ] **Charts remain loading after failure (medium)** — [usePortfolioHistoryCoordinator.ts:23](/home/dev/raaaws/yearn.fi/apps/yearn/src/components/pages/portfolio/hooks/usePortfolioHistoryCoordinator.ts:23): an initial failed history request leaves charts showing a loader instead of the error view.
  Done when: initial failures and timeouts show errors on Balance, Growth, and Annualized, with regression coverage preserving successful and pending states.
  Provenance: `1bcbad968`.

## Verdict

COMMENT

---

## How This Was Reviewed

Reviewed with the [review-pr-workflow skill](https://github.com/yearn/webops/blob/main/skills/review-pr-workflow/SKILL.md): five lenses, each finding independently verified by native session agents using the same model. One candidate was refuted and dropped.

---

Tier: full — 93 files, financial calculations and sensitive configuration.
Verify-agent: session; run-checks: true.
Refuted: 1; discarded: 0; advisories: 0; duplicates merged: 1; unverified: 0.

Lint, type checking, and **969 tests passed**. Browser checks covered desktop/mobile, USD/ETH/Index, 90D/ALL, sorting, and tooltips. The failed-request loader was reproduced.

Remaining coverage gaps:

- Clean production build was not rerun.
- External API consumer migration and deployment credentials remain unconfirmed.
- Real Redis cold/warm/invalidation lifecycle remains untested.
- Empty-wallet and sub-dollar/mixed-sign browser cases remain untested.

w84april:
@rossgalloway thx! `Stale USD Growth pricing` issue made sense, fixed in https://github.com/yearn/yearn.fi/pull/1385/commits/4966533a94bf3fd774c855ecf6ea57dd0d905017
2nd seems irrelevant to me, 3rd was also fixed in https://github.com/yearn/yearn.fi/commit/4966533a94bf3fd774c855ecf6ea57dd0d905017

rossgalloway:
> 2nd seems irrelevant to me

I kept getting this is automated reviews so I asked about it. The main issue seems to be that if 2 vaults happen to have offsetting growth contributions (i.e. +$5 and -$5) over a particular timeframe, they will be dropped as components of the index chart because their offsetting values leads to their individual contributions each reading as 0.

I pushed a fix for this in commit [1691d3d](https://github.com/yearn/yearn.fi/pull/1385/commits/1691d3d855c82e9caf356944a3873d516f95a514)

rossgalloway (commented):
- [ ] **Missing Growth history (high)** — Some multi-token portfolios lose USD/Index history despite complete upstream prices: combining current and historical requests changes timestamp keys, and the receipt-price selector drops valid prices. [pnlSimple.ts:685](https://github.com/yearn/yearn.fi/blob/1691d3d855c82e9caf356944a3873d516f95a514/apps/yearn/src/server/lib/holdings/services/pnlSimple.ts#L685)
  Done when: Two-token portfolios with deposits on different days retain their history when current and receipt prices are fetched together.
  Provenance: dc3be54df.

- [ ] **Unbounded price-per-share downloads (medium)** — Cold portfolio loads can exceed the twelve-request limit because slots are released before response bodies finish downloading; a controlled reproduction reached 26 simultaneous downloads. [kong.ts:202](https://github.com/yearn/yearn.fi/blob/1691d3d855c82e9caf356944a3873d516f95a514/apps/yearn/src/server/lib/holdings/services/kong.ts#L202)
  Done when: Delayed bodies and concurrent callers remain within twelve outstanding downloads, and body-read failures release capacity.
  Provenance: 1bcbad968.

**Verdict: COMMENT**

---

Reviewed with the [review-pr-workflow skill](https://github.com/yearn/webops/blob/main/skills/review-pr-workflow/SKILL.md): five review lenses, each published finding verified by a fresh native session agent using the same model. Two candidate findings were refuted and dropped.

rossgalloway:
> * [ ]  **Missing Growth history (high)** — Some multi-token portfolios lose USD/Index history despite complete upstream prices: combining current and historical requests changes timestamp keys, and the receipt-price selector drops valid prices. [pnlSimple.ts:685](https://github.com/yearn/yearn.fi/blob/1691d3d855c82e9caf356944a3873d516f95a514/apps/yearn/src/server/lib/holdings/services/pnlSimple.ts#L685)
>       Done when: Two-token portfolios with deposits on different days retain their history when current and receipt prices are fetched together.
>       Provenance: [dc3be54](https://github.com/yearn/yearn.fi/commit/dc3be54dfc6f01fd9a8e8221789dcc0ea7334b80).

## Missing Growth history despite complete upstream prices — confirmed, high severity

Validated on `feat/ledger-monorepo` at `1691d3d855c82e9caf356944a3873d516f95a514`.

Complete historical price responses can produce `missing_receipt_price` for every vault, leaving USD and Index Growth history entirely `null`. The failure occurs between price-request normalization and receipt-price selection.

The finding is valid, but the original two-deposit example needs adjustment: A deposited Jan 2 and B deposited Jan 3, with current prices on Jan 4, retained USD/Index history in testing. The precise trigger is the shared contiguous-range optimization. A single-token portfolio can also trigger it.

### Root cause

All service paths below are relative to `apps/yearn/src/server/lib/holdings/services/`.

1. `pnlSimple.ts:557–561` creates receipt-price requests at UTC midnight, including the following midnight when that day is settled.
2. `pnlSimple.ts:2942–2948` combines receipt requests with current/end-of-period price requests.
3. `prices.ts:119–135` detects when every token has the same contiguous set of normalized UTC days, containing at least two days.
4. `prices.ts:702–710` replaces the original requested timestamps with normalized end-of-day timestamps for the optimized range requests.
5. `prices.ts:763–770` materializes the returned prices against those replacement timestamps.
6. `pnlSimple.ts:685–686` selects receipt prices using exact original timestamp keys. The original midnight keys are absent, so valid prices are discarded.
7. The PnL engine excludes affected vaults from USD/Index history.

The exact-match selector was introduced by `dc3be54df`, which is an ancestor of the reviewed head. The range optimization itself predates that commit.

### Reproduction

Use two distinct underlying tokens with complete daily prices and complete vault PPS:

- Token A: deposit Jan 2.
- Token B: deposits Jan 2 and Jan 3.
- Deposits occur at noon UTC.
- Latest settled/current price timestamp: Jan 4 at 23:59:59 UTC.

| Stage | Token A | Token B |
|---|---|---|
| Required receipt keys | Jan 2 and Jan 3 at 00:00 | Jan 2, Jan 3 and Jan 4 at 00:00 |
| Receipt-only fetch result | Both original keys retained | All three original keys retained |
| Add Jan 4 current price | Normalized days become Jan 2–4 | Normalized days remain Jan 2–4 |
| Combined fetch result | Jan 2–4 at 23:59:59 | Jan 2–4 at 23:59:59 |
| Exact receipt selection | 0 of 2 required keys retained | 0 of 3 required keys retained |

Adding the current timestamp makes both normalized ranges identical, enabling the optimization.

The combined asset fetch returns six valid daily prices with zero failed batches. Nevertheless, all five required receipt keys are lost.

### Observed downstream result

With receipt-only results, the real history calculator produces populated USD/Index history. For the synthetic fixture, the final values are:

- `growthWeightUsd`: `8`
- `growthIndex`: `101.98019801980197`

Through the combined portfolio orchestration:

- Both vaults receive `missing_receipt_price`.
- `completeVaults` is `0`; `partialVaults` is `2`.
- Every history point has `growthWeightUsd: null` and `growthIndex: null`.
- `familySeries` is empty.
- `growthIsPartial.usd` and `growthIsPartial.index` are both `true`.
- Current-price Growth rows remain populated.

The frontend consumes these fields in `PortfolioHistoryChart.tsx:610,628` and displays the incomplete-valuation state at `:981–990`.

The incomplete result also reaches the cache writer because no upstream fetch failed.

### Controls

| Scenario | USD/Index history |
|---|---|
| A Jan 2; B Jan 3; current Jan 4 | Available |
| A Jan 2; B Jan 2 and Jan 3; current Jan 4 | Entirely null |
| Single token deposited Jan 2; current Jan 4 | Entirely null |
| A Jan 2; B Jan 2 and Jan 3; current Jan 6 | Available |

The Jan 6 control leaves a gap in the requested range and avoids the shared contiguous-range optimization.

### Acceptance criteria

- The failing fixture above retains every required receipt key when receipt and current prices are fetched together.
- Receipt-only and combined fetching resolve each receipt to the same intended price.
- The portfolio calculation retains numeric USD/Index history and both family series.
- Complete upstream data does not produce `missing_receipt_price`.
- Coverage includes single-token portfolios, differing and identical date ranges, shared-range window splitting, and ordinary batch requests.
- Tests use varying daily prices to verify the selected observation, rather than only asserting map size.
- Genuinely missing receipt prices remain unavailable; current or exit prices must not silently substitute for them.

### Validation evidence and limits

- Existing focused suites: **101 tests passed** across `prices.test.ts`, `pnlSimpleHistory.test.ts`, and `pnlSimple.test.ts`.
- Isolated reproduction and controls: **5 checks passed**, asserting the observed behavior.
- Enabling the desired acceptance assertion fails against the unchanged branch:
  `USD Growth must remain populated with complete prices: expected null not to be null`.

The reproduction uses the real price fetcher, selection, valuation, receipt accounting, and portfolio-history calculation. HTTP price responses, wallet/PPS acquisition, and cache storage use controlled fixtures.

This validates the failure path with complete upstream data. It does not establish prevalence across live wallets. Application source was not changed.

PR #1389, opened 2026-09-04
Fix staked yBOLD Enso withdrawal amounts

rossgalloway:
### Summary
Staked yBOLD to USDC zaps sent a yBOLD asset amount to Enso as if it were a ysyBOLD share amount. This could ask the wallet for more ysyBOLD than it held and make transaction simulation fail.

The withdrawal flow now converts partial asset amounts with ERC-4626 `previewWithdraw`, uses the redeemable share balance for MAX, and keeps approval and quote state aligned with the real source-token amount.

### How to review
Start with `useWithdrawFlow.ts` and `ensoWithdrawAmount.ts`. Check that partial staking withdrawals wait for the wrapper share preview, MAX withdrawals use the full redeemable staking balance, and vault-share withdrawals keep their existing amount.

Then review `useEnsoWithdraw.ts` and `index.tsx` to confirm that route loading, errors, approvals, and protected quote state all use the converted source amount. The added tests cover the original yBOLD values, MAX, non-staking withdrawals, legacy fallback, and preview failure.

### Test plan
- [x] Manual: Reproduced the original `transferFrom` underflow and confirmed that a corrected Enso transaction succeeds in Ethereum `eth_call`.
- [x] Automated: `bun run check:widget` (310 tests)
- [x] Automated: `bun run tslint:ybold`
- [x] Automated: `bun run tslint:yearn`
- [x] Automated: `bun run test:ybold` (5 tests)
- [x] Automated: `bun run test:yearn` (820 tests)
- [x] Automated: Next.js webpack production build, using the container-only TypeScript CLI workaround

### Risk / impact
This changes fund-moving quote and approval amounts for Enso withdrawals from staking wrappers. Direct vault withdrawals are unchanged. Known wrappers fail closed if the share preview cannot be read, while legacy adapters keep the existing one-to-one fallback. Revert commit `416e08e96` to roll back the change.


w84april (approved):
lgtm

PR #1392, opened 2026-09-09
Codex/restore stacked growth attribution 2

w84april:
### Summary

Keep portfolio Index charts available when some vaults have incomplete valuation data, using a fixed subset of reliably valued vaults.

- Keep Index contributions aligned with the included subset and invalidate outdated history caches.
- Align partial-data warnings across USD, ETH, and Index without showing excluded-vault counts.
- Restore subtler chart colors and improve contribution stacking order.

### Test plan

Type checks, lint, and portfolio regression tests passed. Manually check partial-history wallets across 1Y/ALL and USD/ETH/Index views.

### Risk / impact

Index now represents the reliably valued subset when data is incomplete. No transaction-flow changes.

### DIFF
--- a/apps/yearn/src/components/pages/portfolio/components/PortfolioGrowthContributionsChart.tsx
+++ b/apps/yearn/src/components/pages/portfolio/components/PortfolioGrowthContributionsChart.tsx
@@ -20,7 +20,6 @@ import {
   getChartMonthlyTicks,
   getChartWeeklyTicks
 } from '@pages/vaults/utils/charts'
-import { LIGHT_MODE_COLORS } from '@shared/components/AllocationChart'
 import { formatUSD } from '@shared/utils'
 import type { ReactElement } from 'react'
 import { useMemo } from 'react'
@@ -46,10 +45,19 @@ type TPresentedContributionSeries = TPortfolioGrowthContributionSeries & {
 }
 
 const MAX_VAULTS = 8
-const CONTRIBUTION_COLORS = LIGHT_MODE_COLORS
-const OTHER_COLOR = '#d7e6ff'
-const INDEX_BASE_COLOR = '#e8f1ff'
-const TOTAL_COLOR = '#0657f9'
+const CONTRIBUTION_COLORS = [
+  '#46a2ff',
+  '#7bb3a8',
+  '#e1a23b',
+  '#b67ae5',
+  '#f472b6',
+  '#f97316',
+  '#14b8a6',
+  '#94adf2'
+] as const
+const OTHER_COLOR = '#94a3b8'
+const INDEX_BASE_COLOR = '#80b7f4'
+const TOTAL_COLOR = '#2578ff'
 const LINE_HEADROOM = 1.05
 const CHART_MARGIN = {
   ...CHART_WITH_AXES_MARGIN,
@@ -353,7 +361,7 @@ export function PortfolioGrowthContributionsChart({
             stroke={item.color}
             strokeWidth={0.75}
             fill={item.color}
-            fillOpacity={item.isBase ? 0.28 : item.isOther ? 0.3 : 0.45}
+            fillOpacity={0.1}
             connectNulls={mode !== 'eth'}
             tooltipType={'none'}
             isAnimationActive={false}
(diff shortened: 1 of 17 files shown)


rossgalloway:
had a long chat with codex about this branch. the response below is pretty long but hopefully legible. Happy to take over the changes and push to this branch if that is an easier solution than the back and forth.

---

<html>
<body>
<html><head></head><body><p>This review recommends keeping the existing USD, ETH, and Index calculations, while making their handling of missing data more consistent.</p><h3>What the charts measure</h3><p>All three charts start from the same position history: deposits, withdrawals, shares held, and vault price per share (PPS). PPS already reflects the vault’s compounding.</p><p>The charts use that history to answer different questions:</p><ul><li><strong>USD growth:</strong> How much underlying growth did these positions earn, valued using each deposit lot’s receipt-time USD price?</li><li><strong>ETH growth:</strong> The same type of growth amount, valued using each lot’s receipt-time ETH price.</li><li><strong>Index:</strong> How did the portfolio perform relative to the capital invested during each interval, adjusting for deposits and withdrawals?</li></ul><p>USD and ETH still need deposit and withdrawal accounting to separate earnings from principal. The Index additionally uses participating capital as the denominator when calculating returns.</p><p>For example, earning $10 on $100 and earning $10 on $1,000 produce the same USD growth amount, but different returns. This is why Index is a separate calculation, rather than another display unit.</p><h3>How the architecture currently works</h3><p>The backend builds position ledgers from wallet events. These ledgers retain deposit lots, remaining shares, principal, receipt prices, and realized growth.</p><p>The valuation code then uses PPS to convert remaining shares into underlying assets and calculate unrealized growth.</p><p>From there, the paths differ:</p><ul><li>USD growth is calculated when the vaults are materialized, then summed.</li><li>ETH growth has a separate calculation over the ledgers, then sums the available values.</li><li>Index uses USD-weighted growth and capital. This PR adds a fixed set of excluded vaults and rebuilds history when that set changes.</li></ul><p>The frontend receives the totals and per-vault series. It derives warnings from a mixture of final summary status, Index exclusions, and missing values in the displayed series.</p><p>The problem is that these paths do not share one consistent representation of unavailable valuation data. Some missing inputs become numeric values, while others become <code dir="ltr">null</code> or cause exclusion.</p><h3>Confirmed problem: missing PPS can appear as a loss</h3><p>For an open position, the calculation should be:</p><pre dir="ltr"><code>current underlying = shares × PPS
unrealized growth = current underlying − remaining principal</code></pre><p>If PPS is unavailable, current underlying is also unavailable.</p><p>Currently, both USD and ETH paths can substitute zero for current underlying and then subtract principal. This makes an unavailable position look worthless.</p><p>A focused test reproduced the issue with 100 deposited units:</p><div><div><div>
Point | USD growth | ETH growth at $2/ETH
-- | -- | --
Initial PPS of 1 | 0 | 0
PPS unavailable | −100 | −50
PPS recovered to 1.1 | +10 | +5

</div></div></div><p>This gives USD and ETH a shared completeness policy, with differences only where their price requirements differ.</p><p><strong>3. Keep the Index’s fixed-subset behavior.</strong></p><p>Index needs growth and capital from the same set of vaults. Omitting a vault’s growth while retaining its principal would understate returns. Changing membership during the series can also distort linked returns.</p><p>Keep the exclusion of both growth and capital, the history rebuild, and the alignment of stacked Index contributions with the included subset.</p><p><strong>4. Return completeness information with each historical value.</strong></p><p>Each output should identify whether it is complete, partial, estimated, or unavailable, and which contributions are missing.</p><p>Calculate this information before selecting vaults for display. The frontend should not infer portfolio completeness from a filtered list of chart series or only from the latest valuation.</p><p>USD and ETH may still show known contributions as a partial total. However, the chart should identify changes in coverage and avoid presenting them as uninterrupted investment performance.</p><p><strong>5. Attach estimate flags to the value being plotted.</strong></p><p>The chart plots <code dir="ltr">growthWeightUsd</code>, but currently attaches <code dir="ltr">growthUsdEstimated</code>. These fields describe different valuations.</p><p>Estimate and completeness flags should describe the actual plotted field. A missing exit price should not label a receipt-weighted value estimated when all its required inputs are available.</p><h3>Rationale and priority</h3><p>These changes preserve the purpose of each chart. They make the handling of uncertainty consistent: <strong>missing information must not appear as a financial loss, a genuine zero, or a complete portfolio total.</strong></p><p>The first priority is the confirmed missing-PPS loss bug. The Index protection should remain. Completeness metadata and estimate labels should then be aligned with the actual calculations.</p><p>Regression tests should cover unavailable PPS and recovery, missing conversion prices, partially available portfolios, and fully exited positions. Existing focused tests pass, but the additional missing-PPS test exposed a gap in the current coverage.</p></body></html>
</body>
</html>

Issue #474, opened 2024-01-05
Bug: unable to claim veYFI boost rewards


defiglenn:
### What happened 🆘

Unable to claim veYFI boost rewards in the form of dYFI. Spinning wheel.
![IMG_20240105_164617_403](https://github.com/yearn/yearn.fi/assets/86468557/da3ddc53-0999-4f9c-bbaf-5b6a93ff1c9e)

Normal issues to fix such as using a different browser, or clearing cache, not helped.

### To Reproduce 🔂

DM me for user's wallet address

### More 🔎
- Wallet: ledger
- Browser : safari and chrome

PR #1397, opened 2026-09-14
AIO Improvements

w84april:
## Summary

Improve how crawlers and other tools discover public vault data, clarify freshness timestamps, and give users a dedicated view of service availability.

### Vault discovery and freshness

- Correct canonical URLs for Home and Vaults, and advertise the vault catalog’s Markdown endpoint.
- Include eligible public vaults in the sitemap, validate and deduplicate entries, and return explicit errors when upstream data is unavailable.
- Use available upstream timestamps for sitemap `lastmod`; remove generated timestamps, `priority`, and `changefreq`. This follows [Google’s sitemap guidance](https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap).
- Validate Markdown responses, bound upstream requests, and distinguish response generation time, source update time, and Kong cache refresh time.
- Consume the `x-last-refresh` header introduced in [Kong #474](https://github.com/yearn/kong/pull/474). Missing or invalid values remain unreported rather than receiving an inferred timestamp.

Public routes remain accessible to crawlers, consistent with [OpenAI’s crawler documentation](https://developers.openai.com/api/docs/bots). Markdown and `llms.txt` provide convenient discovery formats for consuming tools; [Google does not require special AI markup or use `llms.txt` for Search](https://developers.google.com/search/docs/fundamentals/ai-optimization-guide).

### System status

- Add a server-rendered `/status` page and public `/api/status` endpoint covering Kong, Yearn Prices, portfolio activity, routing, CMS, token assets, and configured network RPCs.
- Render initial status content on the server so it is available before client-side JavaScript runs, following [Google’s rendering guidance](https://developers.google.com/search/docs/crawling-indexing/javascript/javascript-seo-basics).
- Add desktop and mobile status indicators, with hover, keyboard-focus, and reduced-motion support.
- Align the status page with the shared content width and **Home → Vaults → Status** navigation.

## How to review

1. Inspect canonical URLs, sitemap entries, and catalog/detail Markdown responses.
2. Check `/status` and `/api/status`, including unavailable services and missing Kong refresh headers.
3. On mobile, open the menu from `/vaults` and select **Full system status**. Confirm `/status` appears with the menu closed.
4. Check the desktop indicator’s hover/focus behavior and status-page alignment.

## Test plan

- [x] 48 focused tests across eight files.
- [x] Type checking and lint checks.
- [x] Mobile navigation verified in a browser at 390 × 844.

### DIFF
--- a/apps/yearn/app/api/status/route.ts
+++ b/apps/yearn/app/api/status/route.ts
@@ -1,0 +1,4 @@
+export const runtime = 'nodejs'
+export const dynamic = 'force-dynamic'
+
+export { GET, OPTIONS } from '@/server/status'
--- a/apps/yearn/next.config.ts
+++ b/apps/yearn/next.config.ts
@@ -4,6 +4,7 @@ import type { NextConfig } from 'next'
 
 const WORKSPACE_ROOT = fileURLToPath(new URL('../..', import.meta.url))
 loadEnvConfig(WORKSPACE_ROOT, process.env.NODE_ENV !== 'production', console, true)
+const siteUpdatedAt = process.env.NEXT_PUBLIC_SITE_UPDATED_AT || new Date().toISOString()
 
 const CSP_REPORT_URI =
   'https://o4510960324837376.ingest.us.sentry.io/api/4510960614375424/security/?sentry_key=6b1b2932f1532eff2227d01a122adbb4'
@@ -51,6 +52,9 @@ const nextConfig: NextConfig = {
   reactStrictMode: true,
   poweredByHeader: false,
   transpilePackages: ['@yearn/vault-widget'],
+  env: {
+    NEXT_PUBLIC_SITE_UPDATED_AT: siteUpdatedAt
+  },
   turbopack: {
     resolveAlias: {
       '@safe-global/safe-apps-sdk': '../../node_modules/@safe-global/safe-apps-sdk/dist/esm'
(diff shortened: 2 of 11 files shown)


rossgalloway:
Manual Review:

Some services seemed to go down briefly when I was testing and they showed up as unavailable. ENSO had some issues and I couldn't get a quote but then when it said it was good, I could. So it seems to be working. But CMS and token assets were also showing as down, even though they weren't. Not sure what the issue was but it is something we should keep an eye on.

One real issue I noticed was that the header state/dot got out of sync with the status page state/dots. Seems like we may need to be a bit more defensive to make sure this can't happen.
Screenshot 2026-09-18 152107

A small fix is that we should change the breadcrumbs so status isn't nested in vaults, but is a top level element. It should simply show `Home > Status`.
Screenshot 2026-09-18 151935

otherwise, the status page looks nice and the dot in the header is clean. I would maybe explore adding a bit more space between the dot and the "vaults" button, so it matches the spaces between the other elements. 
image

automated review coming shortly.

rossgalloway (commented):
## Issues

- [ ] **Stale status page (medium)** — After an outage or recovery, `/status` can contradict the header because its service rows and check time never refresh while the header polls every minute. [page.tsx:60](https://github.com/yearn/yearn.fi/blob/3528f06eb210e189ec2e21ebf19133312b9b3dd8/apps/yearn/app/%28wallet%29/status/page.tsx#L60)  
  Done when: Outage and recovery update page details and desktop/mobile indicators consistently without reloading, while preserving server-rendered initial content.  
  Provenance: `c0e77efb`.

## How This Was Reviewed

5 review lenses, each published finding verified; defects by session agents (same model, independent contexts), advisory version facts by native session agents. 0 candidate findings were refuted and dropped.

